PR #1, opened 2024-11-21
chore: set up initial actions & docs

JakobJingleheimer:


bmuenzenmeyer:
just throwing this previous conversation here https://github.com/nodejs/release-cloudflare-worker/issues/132 - it's not a blocker, but i honestly dont know how to manage this sort of fragmentation across the project - perhaps it doesnt matter

package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "userland-migrations",
+  "version": "1.0.0",
+  "description": "A collection of migration recipes for userland code.",
+  "main": "index.js",
```
AugustinMauroy, line 5:
```suggestion
```

JakobJingleheimer:
I think some fragmentation is fine. I think it's more important to keep standards similar. How they are enforced is an implementation detail, and the existing modus operandi of the project is, without a compelling reason to the contrary, that's author's choice.

ESLint being "in the family" is maybe a compelling reason, but i think not a clear-cut one.

I don't have much preference though. ESLint has been fine when I've used it before. I only tossed in biome here because of the reasons Augustin mentioned.

AugustinMauroy:
Note, you put the MIT license in `package.json` but not on the repo

package.json
```diff
@@ -0,0 +1,35 @@
...
+  "scripts": {
+    "test:e2e": "node --no-warnings --experimental-test-coverage --test-reporter=lcov --test-reporter-destination=./coverage.lcov --test-reporter=spec --test-reporter-destination=stdout --import './build/snapshots.ts' -test --test-coverage-include='recipes/**/*' --test-coverage-exclude='**/*.e2e.m(j|t)s' './packages/*/*.e2e.m(j|t)s'",
+    "test:lint": "biome lint ./recipes",
+    "test:types": "tsc",
```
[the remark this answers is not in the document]
JakobJingleheimer: No emit is already in tsconfig 🙂

If you feel strongly, sure. I think the consistency (`test:…`) is better though.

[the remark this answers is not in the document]
JakobJingleheimer: 🤔 I do generally think of type checking as a test suite. Linting is conformity but doesn't verify whether the code will work. Types and tests _do_ verify the code will work: if a type-check fails, that _should_ mean the code will fail under some condition (just like a test-case).

CONTRIBUTING.md
```diff
@@ -0,0 +1,18 @@
+# Contributing
```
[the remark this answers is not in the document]
JakobJingleheimer: Good idea!

This PR isn't finished btw. Just as much as I could get done last night. But it's ready for input/discussion 🙂

biome.jsonc
AugustinMauroy, line 1: maybe use `biome format --write .` and specify file here and include json (this file x'D)

JakobJingleheimer: Biome is reading this file; the (immediate) problem is that the config for `useImportType` is specified wrong.

JakobJingleheimer: Thank you 🙇‍♂️

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "editor.formatOnSave": true,
+    "editor.defaultFormatter": "biomejs.biome",
+    "javascript.updateImportsOnFileMove.enabled": "always",
+    "typescript.updateImportsOnFileMove.enabled": "always"
```
JakobJingleheimer, line 5: I think we need to set an additional property so that VS Code doesn't change `d.ts` → `.js` when it updates imports. I don't recall what the config prop is though.

CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+
+New recipes are added under `./recipes` in their own folder, succinctly named for what it does. General-purpose recipes have simple names like `correct-ts-specifiers`. A suite of migrations has a name like `migrate from 18 to 20`, and more specific migrations are named like `migrate fs.readFile from 18 to 20`.
+
+
+## Before pushign a commit
+
+- Do `node --run lint:fix` to fix linting issues
+- Do `node --run format:fix` to format code
+- Do `node --run test:types` to check types
+- Do `node --run test` to run tests (it's should pass 😃)
```
JakobJingleheimer, line 26: Should we create a superset of these so users just have to do 1 thing? Ex `node --run pre-commit`

AugustinMauroy:
Should we have commit hooks ?

CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+
+CI will run lint & type checking and all included test files against all PRs.
+
+> [!INFO]
+> snapshots will be generated with the file extension `.snap.cjs`.
+
+New recipes are added under `./recipes` in their own folder, succinctly named for what it does. General-purpose recipes have simple names like `correct-ts-specifiers`. A suite of migrations has a name like `migrate from 18 to 20`, and more specific migrations are named like `migrate fs.readFile from 18 to 20`.
+
+
+## Before pushign a commit
```
JakobJingleheimer, line 21:
```suggestion
## Before pushing a commit
```

JakobJingleheimer, line 21:
```suggestion
## Before pushing a commit
```

package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "userland-migrations",
+  "version": "1.0.0",
+  "description": "A collection of migration recipes for userland code.",
+  "scripts": {
+    "test:e2e": "node --no-warnings --experimental-test-coverage --test-reporter=lcov --test-reporter-destination=./coverage.lcov --test-reporter=spec --test-reporter-destination=stdout --import './build/snapshots.ts' --test --test-coverage-include='recipes/**/*' --test-coverage-exclude='**/*.e2e.m(j|t)s' './packages/*/*.e2e.m(j|t)s'",
+    "lint": "biome lint ./",
+    "lint:fix": "biome lint --fix ./",
+    "format": "biome format ./",
```
JakobJingleheimer, line 9: What's the difference between lint and format?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
...
+    strategy:
+      matrix:
+        node-version: [23.x]
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'npm'
+    - run: npm ci
+    - run: npm run lint
```
AugustinMauroy, line 24: this should be done in sub steps ? I'm no sure.
@bmuenzenmeyer is more of an expert on this than I am

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Node.js ${{ matrix.node.version }}
+      uses: actions/setup-node@v4
+      with:
+        node-version: ${{ matrix.node.version }}
+        cache: 'npm'
+    - run: npm ci
+    - run: npm run test:unit
+    - run: npm run test:e2e
```
JakobJingleheimer, line 51: Should these be switched to `node --run …`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
...
+      matrix:
+        node-version: [23.x]
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'npm'
+    - run: npm ci
+    - run: npm run lint
+    - run: npm run test:types
```
JakobJingleheimer, line 25: Same: Should these be switched to `node --run …`?

Also, should we add format?

AugustinMauroy:
```
error TS6059: File '/home/runner/work/userland-migrations/userland-migrations/build/snapshots.ts' is not under 'rootDir' '/home/runner/work/userland-migrations/userland-migrations/recipes'. 'rootDir' is expected to contain all source files.
```

should we include or exclude this file ?

.github/workflows/ci.yml
AugustinMauroy: maybe

package.json
AugustinMauroy: if you take look to biome rules set there are two part linter and formater
- linter is a code analyser 
- formater is a code formater 

IDK if you get the idea put it's two separate things

JakobJingleheimer:
> Should we have commit hooks ?

I think no: those can get hella annoying. If a user wants to commit code that will fail, that doesn't hurt anything, so let's not get in the way—maybe they have a good reason (like sharing something incomplete).

.vscode/settings.json
AugustinMauroy: I've no idea, but I think this behaviour is gone now, but not for sure.

JakobJingleheimer:
> should we include or exclude this file ?

I think include

.vscode/settings.json
JakobJingleheimer: Definitely not gone. It happened to me like yesterday.

JakobJingleheimer:
Hmm, I'm not sure why this works fine locally but fails in CI

```
> userland-migrations@1.0.0 test:unit
> node --no-warnings --experimental-test-coverage --test-reporter=lcov --test-reporter-destination=./coverage.lcov --test-reporter=spec --test-reporter-destination=stdout --experimental-test-module-mocks --import './build/snapshots.ts' --test --test-coverage-include='recipes/**/*' --test-coverage-exclude='**/*.spec.m(j|t)s' --test-coverage-lines=0.8 './packages/*/*.spec.m(j|t)s'

node: bad option: --test-coverage-include=recipes/**/*
node: bad option: --test-coverage-exclude=**/*.spec.m(j|t)s
node: bad option: --test-coverage-lines=0.[8]
```

https://github.com/nodejs/userland-migrations/actions/runs/11986924657/job/33420496685?pr=1#step:5:9

package.json
```diff
@@ -3,11 +3,12 @@
...
   "scripts": {
-    "test:e2e": "node --no-warnings --experimental-test-coverage --test-reporter=lcov --test-reporter-destination=./coverage.lcov --test-reporter=spec --test-reporter-destination=stdout --import './build/snapshots.ts' --test --test-coverage-include='recipes/**/*' --test-coverage-exclude='**/*.e2e.m(j|t)s' './packages/*/*.e2e.m(j|t)s'",
-    "lint": "biome lint ./",
-    "lint:fix": "biome lint --fix ./",
-    "format": "biome format ./",
     "format:fix": "biome format --fix ./",
+    "format": "biome format ./",
+    "lint:fix": "biome lint --fix ./",
+    "lint": "biome lint ./",
+    "pre-commit": "node --run lint:fix; node --run format:fix; node --run test:types; node --run test:unit; node --run test:e2e",
```
AugustinMauroy, line 10: why not use `&&` instead of `;` to a have a "strict" chain if one fail rest will not follow

JakobJingleheimer: I considered that. I did it this way because they're independent, and allowing others to continue at least gets it closer (and perhaps the failures can be fixed manually).

JakobJingleheimer:
@bmuenzenmeyer are you okay with biome, or do you have a strong preference for eslint?

bmuenzenmeyer:
I'm like a 4 outta 10 in these tertiary repos.

JakobJingleheimer:
> I'm like a 4 outta 10 in these tertiary repos.
> 
> I'm not a blocker.

Gotcha :) Just wanted to ensure it isn't "Jacob's way or the highway".

bmuenzenmeyer:
> Hmm, I'm not sure why this works fine locally but fails in CI

it's only complaining about flags written after `--test`... but it's odd that the failing CI doesn't seem to exhibit a pattern across os or version.

bmuenzenmeyer:
note 

- all windows builds that are failing have this error message `'t)s'' is not recognized as an internal or external command,
`
- all macos or ubuntu failures have the `bad option` messages

AugustinMauroy:
maybe because newest version use `fs.glob`  or something  like that ?

JakobJingleheimer:
I think it's not [fs.glob](https://nodejs.org/docs/v22.11.0/api/fs.html#fsglobpattern-options-callback) itself (available since 22.0.0, and 23.x is failing).

Putting the quotes around the values ensures they are processed by node (should be consistent behaviour); without quotes, the shell handles the globs (inconsistent behaviour).

What could be different between local and CI? It's not OS, because CI's macos is also failing.

I wonder if it's something like double vs single quotes?

richardlau:
Tests on Node.js 20 are failing because https://github.com/nodejs/node/pull/53553 is not in Node.js 20.

JakobJingleheimer:
Looks like it was indeed the overly fancy glob—for whatever reason, CI does not like `m(j|t)s` (but doesn't mind `{mjs,mts}`).

AugustinMauroy (approved):
LGMT ! I'm fine as it. maybe open an issue to track v20 support. Should we found a solution (own workflow) or just no support it until we have a codemod need v20.

AugustinMauroy:
BTW if we not support node 20 we can only use `--run` (introduce in 22)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  lint-and-types:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [23.x]
+    steps:
+    - uses: actions/checkout@v4
```
bmuenzenmeyer, line 19: nit: pin these using a sha throughout

see https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

dependabot MIGHT fix this automatically too.

.github/workflows/ci.yml
bmuenzenmeyer: its fine as long as you think linting and types go hand in hand here 👍

.nvmrc
```diff
@@ -0,0 +1 @@
+23
```
bmuenzenmeyer, line 1: do we recommend odd versions? 
at least in my day job, we try to stay on the prod versions. I suppose this IS a tooling author environment 😄

LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) [year] [fullname]
```
bmuenzenmeyer, line 3: what should these values be?

LICENSE.txt
bmuenzenmeyer, line 1: the way i usually see this is no extension

README.md
```diff
@@ -1 +1,25 @@
 # Node.js userland migrations
+
+This repository contains codemodes (automated migrations) for "userland" code. These are intended to facilitate adopting new features and upgrading source-code affected by breaking changes.
+
+[![Formatted with Biome](https://img.shields.io/badge/Formatted_with-Biome-60a5fa?style=flat&logo=biome)](https://biomejs.dev/) [![Linted with Biome](https://img.shields.io/badge/Linted_with-Biome-60a5fa?style=flat&logo=biome)](https://biomejs.dev) [![Checked with Biome](https://img.shields.io/badge/Checked_with-Biome-60a5fa?style=flat&logo=biome)](https://biomejs.dev)
```
bmuenzenmeyer, line 5: nit: i am not a big fan of these. they are bumperstickers, and are often not accessible to screenreaders. YMMV

package.json
```diff
@@ -0,0 +1,38 @@
...
+    "automation",
+    "codemod",
+    "migrations",
+    "node.js"
+  ],
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/nodejs/userland-migrations/issues"
+  },
+  "homepage": "https://nodejs.org/learn/userland-migrations",
```
bmuenzenmeyer, line 29: nice

bmuenzenmeyer (approved):
some comments!

.github/workflows/ci.yml
JakobJingleheimer: Ah, sure!

README.md
JakobJingleheimer: I think these don't add much value—they seem more marketing for biome. @AugustinMauroy you added them. do you feel strongly?

LICENSE.txt
JakobJingleheimer: AFAIK, github supports .md, .txt, and extensionless. Extensionless to me means an executable, and this is actually just plain text. I don't feel strongly, but I think it is better with the `.txt`.

.nvmrc
JakobJingleheimer: > this IS a tooling author environment 😄

That's what I was thinking.

LICENSE.txt
AugustinMauroy: it's can also be nodejs project ?

.github/workflows/ci.yml
JakobJingleheimer: Not especially, but the have the same setup and separating them into their own jobs seemed wasteful—no need to create a fresh environment for each; they don't affect each other.

.nvmrc
AugustinMauroy: may be use LTS for dev and recommend it for prod (cli usage of the tool)

README.md
AugustinMauroy: I just add it like that any strong opinion.

LICENSE.txt
JakobJingleheimer: OpenJS own the copyright for node and its related projects, so I think it needs to be whatever their thing is.

.github/workflows/ci.yml
ljharb: I don’t really think it’s important to do that on GitHub-owned actions tho.

.github/workflows/ci.yml
ljharb: wasteful for who? actions are free.

Tests run in a matrix; linting only ever needs to run once; it’s generally best to split it for CI hygiene.

package.json
```diff
@@ -0,0 +1,38 @@
...
+  "devDependencies": {
+    "@biomejs/biome": "^1.9.4",
+    "@types/node": "^22.9.1",
+    "typescript": "^5.6.3"
```
AugustinMauroy, line 33:
```suggestion
    "typescript": "^5.7.2
```

.nvmrc
JakobJingleheimer: I think these would not be used in prod? @alexweej perhaps your use-case is different than what I'm imagining.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
...
+        cache: 'npm'
+    - run: npm ci
+    - run: npm run lint
+    - run: npm run format
+    - run: npm run test:types
+
+  tests:
+    strategy:
+      fail-fast: false
+      matrix:
```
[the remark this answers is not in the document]
JakobJingleheimer: Oww snap! yes. Um, how do I do that?

README.md
JakobJingleheimer: Sorry, I don't understand—you're saying you _do_ feel strongly?

README.md
```diff
@@ -1 +1,25 @@
...
+```
+
+* `transform` - name of transform. see available transforms below.
+* `path` - directory to transform. defaults to the current directory.
+
+See the [codemod CLI doc](https://go.codemod.com/cli-docs) for a full list of available commands.
+
+## Available codemods
+
+All Node.js codemods are also available in the [Codemod Registry](https://codemod.com/registry?framework=node.js).
```
alexbit-codemod, line 25: lets use: https://codemod.link/nodejs-official

LICENSE.txt
JakobJingleheimer: `LICENSE` it is then :) I'll rename it after Robin chimes in (if I do it before, I think GitHub will hide this thread).

.github/workflows/ci.yml
JakobJingleheimer: Electricity? Spinning up a new environment may not cost us money, but it costs IRL resources—unless I'm misunderstanding what will happen?

README.md
AugustinMauroy: I wanted to say that I just added it like that. So it doesn't matter to me whether it's there or not.

LICENSE.txt
JakobJingleheimer: Got the answer: https://github.com/openjs-foundation/cross-project-council/blob/HEAD/governance/IP_POLICY_GUIDANCE.md#2-copyright-notices

.github/workflows/ci.yml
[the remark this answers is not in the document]
ljharb: https://github.com/nodejs/userland-migrations/pull/3

JakobJingleheimer:
Noooo, why is CI failing on those globs again 😭

.github/dependabot.yml
```diff
@@ -0,0 +1,23 @@
...
+      interval: monthly
+    commit-message:
+      prefix: ci
+
+  - package-ecosystem: npm # See documentation for possible values
+    directories:
+      - / # Location of package manifests
+      - /recipies/*
+    commit-message:
+      prefix: deps
```
AugustinMauroy, line 21:
```suggestion
      prefix: "deps(recipies): "
```

JakobJingleheimer: Maybe we should have 2 separate ones then so we can differentiate between the root repo and recipes (cuz if we apply your suggestion, it will mark changes to the root repo's deps as belonging to a recipe).

PR #3, opened 2024-11-27
[actions] use ljharb/actions/node/matrix

ljharb:
(into #1)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,14 +25,24 @@ jobs:
     - run: npm run format
     - run: npm run test:types
 
+  matrix:
+    runs-on: ubuntu-22.04
+    outputs:
+      latest: ${{ steps.set-matrix.outputs.requireds }}
+    steps:
+      - uses: ljharb/actions/node/matrix@6f6460d952d0fa896c91a56c2e450cb8c9fd1a21 # main
+        id: set-matrix
+        with:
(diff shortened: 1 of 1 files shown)


AugustinMauroy (approved):
LGTM ! BTW that cool action

JakobJingleheimer (approved):
w00t! Thanks for this 🙏

JakobJingleheimer:
@ljharb something in here causes CI to fail on globs (ex https://github.com/nodejs/userland-migrations/actions/runs/12068897635/job/33654997438)

ljharb:
@JakobJingleheimer rather than reverting, it might be more helpful to print out `node --version` in the tests, to make sure it's using the right one? node itself is failing on the options, not the globs, from what i can see.

richardlau:
https://github.com/nodejs/userland-migrations/actions/runs/12068897635/job/33654997438#step:3:9 says it installed Node.js 20.18.1

JakobJingleheimer:
> @JakobJingleheimer rather than reverting, it might be more helpful to print out `node --version` in the tests, to make sure it's using the right one? node itself is failing on the options, not the globs, from what i can see.

Ah, yes. I reverted because I think this is super cool but non-blocking. Since the parent PR is otherwise ready to go, I think we can figure this out and then as a subsequent enhancement. And the other landed, I think this will be easier to test 🙂

PR #6, opened 2024-11-30
chore: improve setup & CI config

JakobJingleheimer:
* ignore snapshot files for linting
* switch tests run to leverage npm's `--workspaces`
* move CI-specific options

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     - run: npm ci
     - run: node --run lint
     - run: node --run format
-    - run: node --run test:types
+    - run: node --run type-check
 
   tests:
     strategy:
@@ -51,8 +51,6 @@ jobs:
       uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # 4.1.0
       with:
         node-version: ${{ matrix.node.version }}
-        env: NODE_OPTIONS="${{ matrix.node.options }}"
         cache: 'npm'
     - run: npm ci
-    - run: node --run test:unit
-    - run: node --run test:e2e
+    - run: node --run test ${{ matrix.node.options }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,9 +22,9 @@ jobs:
         node-version: ${{ matrix.node-version }}
         cache: 'npm'
     - run: npm ci
-    - run: npm run lint
-    - run: npm run format
-    - run: npm run test:types
+    - run: node --run lint
+    - run: node --run format
+    - run: node --run test:types
 
   tests:
     strategy:
@@ -54,5 +54,5 @@ jobs:
         env: NODE_OPTIONS="${{ matrix.node.options }}"
         cache: 'npm'
     - run: npm ci
-    - run: npm run test:unit
-    - run: npm run test:e2e
+    - run: node --run test:unit
+    - run: node --run test:e2e
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        options: >-
+          --test-coverage-lines=0.8
+          --test-reporter-destination=./coverage.lcov
+          --test-reporter=lcov
         node:
           - version: 23.x
           - version: 22.x
@@ -46,6 +50,7 @@ jobs:
       uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # 4.1.0
       with:
         node-version: ${{ matrix.node.version }}
+        env: NODE_OPTIONS=${{ matrix.node.options }}
         cache: 'npm'
     - run: npm ci
     - run: npm run test:unit
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,11 @@ on:
     branches: ["main"]
   pull_request:
     branches: ["main"]
-    if: github.event.pull_request.draft == false
 
 jobs:
   lint-and-types:
+    if: github.event.pull_request.draft == false
+
     runs-on: ubuntu-latest
 
     strategy:
@@ -27,6 +28,8 @@ jobs:
     - run: node --run type-check
 
   tests:
+    if: github.event.pull_request.draft == false
+
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,10 +30,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        options: >-
-          --test-coverage-lines=0.8
-          --test-reporter-destination=./coverage.lcov
-          --test-reporter=lcov
         node:
           - version: 23.x
           - version: 22.x
@@ -53,4 +49,9 @@ jobs:
         node-version: ${{ matrix.node.version }}
         cache: 'npm'
     - run: npm ci
-    - run: node --run test ${{ matrix.node.options }}
+    - run: |
+        node \
+          --run test \
+          --test-coverage-lines=0.8 \
+          --test-reporter-destination=./coverage.lcov \
+          --test-reporter=lcov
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -7,10 +7,11 @@
     "format": "biome format ./",
     "lint:fix": "biome lint --fix ./",
     "lint": "biome lint ./",
+    "fix-commit": "node --run format:fix; node --run lint:fix",
     "pre-commit": "node --run lint:fix; node --run format:fix; node --run test:types; node --run test:unit; node --run test:e2e",
-    "test:e2e": "node --no-warnings --experimental-test-coverage --test-reporter=lcov --test-reporter-destination=./coverage.lcov --test-reporter=spec --test-reporter-destination=stdout --import './build/snapshots.ts' --test --test-coverage-include='recipes/**/*' --test-coverage-exclude='**/*.e2e.{mjs,mts}' './packages/*/*.e2e.{mjs,mts}'",
+    "test:e2e": "NODE_OPTIONS='--no-warnings --experimental-test-coverage --test-reporter=spec --test-reporter-destination=stdout --import=\"./build/snapshots.mts\" --test-coverage-include=\"recipes/**/*\" --test-coverage-exclude=\"**/*.e2e.{mjs,mts}\"' npm run test:e2e --workspaces",
```
JakobJingleheimer, line 11: Node doesn't allow these within `NODE_OPTIONS` (I don't know why—other experimental flags are supported):

* `--experimental-test-coverage`
* `--experimental-test-module-mocks`
* `--experimental-test-snapshots`

Can anyone think of another way to get these passed to the sub-process `npm run` will create?

.github/workflows/ci.yml
```diff
@@ -46,6 +50,7 @@ jobs:
       uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # 4.1.0
       with:
         node-version: ${{ matrix.node.version }}
+        env: NODE_OPTIONS=${{ matrix.node.options }}
```
ljharb, line 53:
```suggestion
        env: NODE_OPTIONS="${{ matrix.node.options }}"
```
considering there's spaces and newlines in there?

package.json
ljharb, line 12: would it make more sense to put these in an env file, and pass only the env file option in all these places instead?

tsconfig.json
```diff
@@ -24,7 +24,7 @@
 
     /* Interop Constraints */
     "verbatimModuleSyntax": true,
-    "allowSyntheticDefaultImports": false,
+    "allowSyntheticDefaultImports": true,
```
ljharb, line 27: +1000, TS's module system is broken without this setting

tsconfig.json
```diff
@@ -40,5 +40,24 @@
...
+    "**/*.fixture.ts",
+    "**/*.mock.mjs",
+    "**/*.e2e.mjs",
+    "**/*.e2e.mts",
+    "**/*.e2e.js",
+    "**/*.e2e.ts",
+    "**/*.spec.mjs",
+    "**/*.spec.mts",
+    "**/*.spec.js",
+    "**/*.spec.ts",
```
JakobJingleheimer, line 61: This is super verbose, and the tsconfig docs specifically say they support full globs, but using `{mjs,mts,js,ts}` silently fails.

.github/workflows/ci.yml
JakobJingleheimer: I'm not sure if the quotes are necessary; as long as they don't hurt, sure? There are no new lines though thanks to the yaml magic of `>-` (which collapses inter-whitespace into a single space character).

package.json
JakobJingleheimer: Yes! But only if we can figure out the problem where options we need are disallowed within `NODE_OPTIONS`.

AugustinMauroy (approved):
LGTM !

AugustinMauroy:
instead of requesting review maybe add code owner ?

.github/workflows/ci.yml
```diff
@@ -46,6 +51,7 @@ jobs:
       uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # 4.1.0
       with:
         node-version: ${{ matrix.node.version }}
+        env: NODE_OPTIONS="${{ matrix.node.options }}"
         cache: 'npm'
     - run: npm ci
     - run: npm run test:unit
```
AugustinMauroy, line 57:
```suggestion
    - run: node --run test:unit
```

now we can use --run because we don't support v20

package.json
JakobJingleheimer: https://github.com/nodejs/node/pull/56114

JakobJingleheimer: https://github.com/nodejs/node/pull/56114

JakobJingleheimer: I think this is obsolete with the new way of orchestrating this

JakobJingleheimer: 🤦‍♂️ there is a much easier way: `npm` supports passing along CLI args by appending `-- --FLAG_HERE`

JakobJingleheimer:
I solved the CLI arg issue 🥳

However, I'm getting a seemingly erroneous failure in CI:

```
Run node --run lint
  node --run lint
  shell: /usr/bin/bash -e {0}
node: bad option: --run
```

biome.jsonc
AugustinMauroy, line 1: rules can be added:
- https://biomejs.dev/linter/rules/use-block-statements/#_top
- https://biomejs.dev/linter/rules/no-yoda-expression/
- https://biomejs.dev/linter/rules/use-block-statements/


Found nothing about https://github.com/nodejs/userland-migrations/pull/7#discussion_r1874573979

.github/workflows/ci.yml
```diff
@@ -10,22 +10,29 @@ on:
...
-    - run: npm run test:types
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # 4.2.2
+      - name: Set up Node.js ${{ matrix.node.version }}
+        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # 4.1.0
+        with:
+          node-version: ${{ matrix.node.version }}
+          cache: 'npm'
+      - run: npm ci
+      - run: node --run lint
+      - run: node --run format
```
AugustinMauroy, line 30: maybe use prebuilt GA by biome because it's include a reporter so on PR we see where there are something wrong.

https://biomejs.dev/recipes/continuous-integration/#github-actions

biome.jsonc
JakobJingleheimer: How do we set options for use-block-statements? (I think we should use ["multi"](https://eslint.org/docs/latest/rules/curly#multi))

.github/workflows/ci.yml
JakobJingleheimer: 🤔 That would make CI and local behave differently. I think we should avoid that.

AugustinMauroy: it's run command "biome ci ./".

Eh wait, this ins't on `scripts` other things can be added

biome.jsonc
AugustinMauroy: > (I think we should use ["multi"](https://eslint.org/docs/latest/rules/curly#multi))

It's sad but I don't see support for an equivalent to `multi`. But I don't find it a problem personally, because I think it's fine to force the `{...}` block whatever the condition.

JakobJingleheimer:
@AugustinMauroy ah, it looks like [biome chose](https://biomejs.dev/formatter/option-philosophy/#existing-options) to specifically disallow configuring certain options and unfortunately chose objectively undesirable ones, such as forcefully collapsing line breaks (their choice bloats diffs and makes it more difficult to see what was actually changed). I _despised_ that about prettier, so I would like to replace Biome with ESLint.

bmuenzenmeyer (approved):

biome.jsonc
```diff
@@ -5,10 +5,11 @@
...
   },
-  "formatter": {
-    "indentStyle": "tab",
-    "lineWidth": 100
-  },
+  // "formatter": {
+  //   "indentStyle": "tab",
+  //   "lineWidth": 100,
+  //   "useEditorconfig": true
+  // },
```
bmuenzenmeyer, line 12: did you mean to delete this?

JakobJingleheimer:
```suggestion
```

PR #10, opened 2024-12-12
chore(ci): limit running tests to only changes to specific targets

JakobJingleheimer:


.github/workflows/ci.yml
AugustinMauroy, line 1: we should include `package.json` and `package-lock.json` to run test when we update dependencies

AugustinMauroy (approved):

PR #11, opened 2024-12-12
chore(ci): add "publish" GHA

JakobJingleheimer:
Todo:

- [ ] Set up repo secret for codemod
- [x] Each recipe being published needs to be bundled and then sent to the codemod registry. Thereafter, the bundle generated by CI can be discarded.


.github/workflows/publish.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: export files
+        run: |
+          echo "Modified files: ${{steps.filter.outputs.codemods_files}}"
+          echo "CODEMOD_FILES=${{steps.filter.outputs.codemods_files}}" >> $GITHUB_ENV
+          echo "Modified status: ${{steps.filter.outputs.codemods}}"
+          echo "CODEMOD_STATUS=${{steps.filter.outputs.codemods}}" >> $GITHUB_ENV
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: node
```
AugustinMauroy, line 40:
```suggestion
          node-version-file: '.nvmrc'
```

we should `.nvmrc`

.github/workflows/publish.yml
```diff
@@ -0,0 +1,61 @@
...
+              DIR=$(dirname "$FILE")
+              echo "Checking codemod: $DIR"
+              cd "$ROOT_DIR/$DIR"
+              npx esbuild --bundle --platform=node --outfile=bundle.js
+              npx codemod publish
+              echo "Codemod published"
+            done
+          else
+            echo "No codemods found"
+          fi
```
AugustinMauroy, line 61: What is we use a external script `.sh` or `.js` to allow dev to "simulate" a publish ?

AugustinMauroy:
`npm i` should fix the CI

.github/workflows/publish.yml
JakobJingleheimer: Hmm?

AugustinMauroy (approved):
LGTM !

JakobJingleheimer:
There are a few issues blocking this:

* bundling via `esbuild` encounters a platform-specific node addon used by ast-grep, which can't be bundled
  * `codemod publish` doesn't have a programmatic API, so it'll need leverage `spawn` 
* `codemod publish` can handle the bundling but doesn't support distributed config, and the bundle it builds errors (`transformer cannot be null`)
  * there's not much documentation around what it's doing (they're working on it)

JakobJingleheimer:
> * the bundle it builds errors (`transformer cannot be null`)

This turned out to be a red-herring: it was actually throwing when it got to `module.register('@nodejs-loaders', import.meta.url)` because when `codemod publish` builds the bundle (also via esbuild), esbuild does not realise there is a dependency it needs to include. https://github.com/evanw/esbuild/issues/4016

recipes/correct-ts-specifiers/esbuild.config.ts
```diff
@@ -0,0 +1,8 @@
+import type { BuildOptions } from 'esbuild';
+
+
+const opts: BuildOptions = {
+	format: 'esm',
+};
+
+export default opts;
```
AugustinMauroy, line 8:
```suggestion
export default {
	format: 'esm',
} as BuildOptions;
```

AugustinMauroy:
Any new here Jacob ?

JakobJingleheimer:
Yes! The convo has been happening in Slack and video chat. It publishes but when run, codemod throws an error citing a fictitious problem (`The transformer cannot be null`). I solved the first thing triggered the fictitious error, but the error was still happening. Aleksy tracked down the new one just this evening and sent me a way to access to real issue.

recipes/correct-ts-specifiers/esbuild.config.ts
JakobJingleheimer: Why? Those do different things

AugustinMauroy: It makes no difference to the final rendering, there's one that's more concise in my opinion.

JakobJingleheimer: There are a few benefits to what I did:

* importing a default does not give you intellisense for what is behind it
* a default export can be missing or different and things won't warn you (just blow up at run-time); whereas importing a named export will trigger multiple warnings
* there is a subtle difference between annotating and asserting: an assertion is a form of override/narrowing. Take the example from [TS: Everyday types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions)
  ```ts
  const myCanvas = document.getElementById("main_canvas") as HTMLCanvasElement;
  ```
  `document.getElementById` returns `HTMLElement | null` (because it doesn't know anything about the element being got, so it turns the broadest type). But let's say we know the element is an `HTMLCanvasElement`, which is a superclass of `HTMLElement` (`class HTMLCanvasElement extends HTMLElement {}`). The `as` says to typescript "ignore most of what you know", whereas the annotation says "add this to what you know". You can see the difference in this quick [ts playground example](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAZiATgWwIwwLwwCYmAV2QFMwoA6AcyKgFEAbI40gIQE8BJbACgHIEQeAShgBDCDAASAFQCyAGQDCIsADcx9RiSgBuAFC7QkWAhQAmAFyTZi5WogamsLLgKPK1B1radeAIxGIQtpAA): `as HTMLCanvasElement` has _replaced_ `HTMLElement | null` with `HTMLCanvasElement`.
    * When declaring your own stuff, it's better to annotate.

AugustinMauroy: Oh wow you learnt me something

JakobJingleheimer:
Superseded by https://github.com/nodejs/userland-migrations/pull/50#pullrequestreview-2754427250

PR #70, opened 2025-05-11
feat(jest-to-node-test-runner): add initial implementation

Ceres6:


recipes/jest-to-node-test-runner/src/workflow.test.ts
```diff
@@ -0,0 +1,20 @@
...
+import { execPath } from "node:process";
+import { readFile } from "node:fs/promises";
+import { describe, it, type TestContext } from "node:test";
+import { fileURLToPath } from "node:url";
+
+import { spawnPromisified } from "../test/spawn-promisified.ts";
+import { workflow } from "./workflow.ts";
+
+describe("workflow", () => {
+	it("should update bad specifiers and ignore good ones", async (t: TestContext) => {
```
AugustinMauroy, line 11: no needed to explicitly write this type because it's provide by type def of `it` function 

```suggestion
	it("should update bad specifiers and ignore good ones", async (t) => {
```

recipes/jest-to-node-test-runner/src/workflow.test.ts
```diff
@@ -0,0 +1,20 @@
...
+		const e2eFixtPath = fileURLToPath(import.meta.resolve("./fixtures/e2e/"));
+		const source = await readFile(resolve(e2eFixtPath, "test.ts"), { encoding: "utf-8" });
+		const result = await workflow(source);
+		const importStatement =
+			'import { describe, it } from "node:test";\nimport { expect } from "expect";\n';
+
+		t.assert.deepStrictEqual(result, `${importStatement}${source.split("\n").slice(1).join("\n")}`);
```
AugustinMauroy, line 18: we can use snapshot https://nodejs.org/docs/v22.15.0/api/test.html#snapshot-testing

recipes/jest-to-node-test-runner/tsconfig.json
AugustinMauroy, line 1: add earsableSyntaxOnly to fit with node.js type stripping

AugustinMauroy:
You should add on your todo for this codemod:
- way to remove jest dep, then verify if `expect` is downloaded 
  that need to have something that check which package manager is used 
- check if un-imporded jest `describe, it` is supported by the codemod

recipes/jest-to-node-test-runner/test/spawn-promisified.ts
JakobJingleheimer, line 1: I'll add this to the project root via a dedicated PR. It will live in `project-root/test/spawn-promisified.ts`.

I'll also copy over https://github.com/nodejs-loaders/nodejs-loaders/blob/main/test/snapshots.mjs to correct the filenames used for snapshot files so they work with syntax highlighting.

recipes/jest-to-node-test-runner/tsconfig.json
JakobJingleheimer: I added it to the project root's tsconfig, so CI will enforce it.

recipes/jest-to-node-test-runner/test/spawn-promisified.ts
JakobJingleheimer: https://github.com/nodejs/userland-migrations/pull/71

recipes/jest-to-node-test-runner/src/workflow.test.ts
Ceres6: TS will complain when I use strict assertions. Maybe it is because now they're legacy and I can use `deepEqual` instead

recipes/jest-to-node-test-runner/src/workflow.test.ts
Ceres6: Yeah, I will use that in the end, but since I'm changing this a lot I didn't want to have to update them all the time yet

recipes/jest-to-node-test-runner/src/workflow.test.ts
JakobJingleheimer: I think something is wrong then: you should indeed not have to add the type (I haven't seen it complain before 🤔)

recipes/jest-to-node-test-runner/src/workflow.test.ts
```diff
@@ -0,0 +1,20 @@
+import { resolve } from "node:path";
```
avivkeller, line 1: I don't think there are enough tests here, we should make sure the behavior works in cases of
- Jest-like globals from a different library, I.e. playwright
- Other Jest features

recipes/jest-to-node-test-runner/src/workflow.ts
```diff
@@ -0,0 +1,40 @@
...
+	const root = ast.root();
+
+	const imports = [
+		"describe",
+		"it",
+		"test",
+		"beforeEach",
+		"afterEach",
+		"beforeAll",
+		"afterAll",
```
avivkeller, line 14: What if the user defined and/or imported these from elsewhere?

recipes/jest-to-node-test-runner/README.md
```diff
@@ -0,0 +1,5 @@
+# Jest to Node Test Runner
+
+This package transforms Jest test files into Node Test Runner test files.
+
+This is a one-and-done process, and the updated source-code should be committed to your version control (eg git); thereafter, source-code import statements should be authored compliant with the ECMAScript (JavaScript) standard.
```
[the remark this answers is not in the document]
JakobJingleheimer: "One-and-done" refers to it being a discrete, one-off process (contrary to a loader that runs every time).

recipes/jest-to-node-test-runner/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "README.md",
+    ".codemodrc.json",
+    "bundle.js"
+  ],
+  "keywords": [
+    "codemod",
+    "esm",
+    "typescript"
+  ],
+  "author": "Carlos Espa",
```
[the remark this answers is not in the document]
AugustinMauroy: It's the authors, the poeple that write the code so not an issue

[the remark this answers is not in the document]
JakobJingleheimer: I think that's contrary to the nature of these: They're akin to NPM modules that we "bless". I think we can derefer to Learn articles for precedent (a non-member can be attributed as "author" of an article).

Anyone can publish a codemod in codemod's registry and can associate it to nodejs (regardless of whether they came via userland-migrations: we don't own codemod, we're just partnered with them). Those published via userland-migrations get a little "official" badge.

recipes/jest-to-node-test-runner/src/workflow.test.ts
JakobJingleheimer: It's a draft 😉

recipes/jest-to-node-test-runner/src/workflow.ts
Ceres6: It's still WIP, I'm trying to cover low hanging fruits first so I can test it with simple projects and then I/we could support projects with less standard usage

recipes/jest-to-node-test/src/workflow.test.ts
```diff
@@ -0,0 +1,537 @@
...
+
+describe("workflow", () => {
+	it("should migrate jest to node:test", async (t) => {
+		const e2eFixtPath = fileURLToPath(import.meta.resolve("./fixtures/e2e/"));
+		const source = await readFile(resolve(e2eFixtPath, "test.ts"), { encoding: "utf-8" });
+		const result = await workflow(source);
+
+		t.assert.snapshot(result);
+	});
+
```
AugustinMauroy, line 16:
```suggestion
it("should be a runnable test", async () => {
  /**
  * 1. run the codemod
  * 2. up a child process
  * 3. assert output code + stdout & stderr
  */
});
```

recipes/jest-to-node-test/src/workflow.test.ts
AugustinMauroy, line 1: in some case you use `ts.assert.something` 
you have to import assert module to do some asserting and only if you need to do some snapshots use `t.assert.snapshot`

AugustinMauroy:
Hey on we are going to use new codemod CLI/approach  

https://codemod.com/blog/new-codemod-cli
https://docs.codemod.com/cli/cli-reference

PR #101, opened 2025-07-01
feat(logger): add codemod name

AugustinMauroy:


JakobJingleheimer (changes requested):
Multiple codemods could be running at the same time as part of a multi-part migration. In that case, the module-level "name" will get clobbered.

AugustinMauroy:
🤦🏼 you have right

avivkeller (approved):
Note that it's `Codemod`, not `Codemode`

AugustinMauroy:
styletext isn't disponible on JSSG

PR #133, opened 2025-07-25
add codemod to enforce 'new' with zlib class(fix #105)

### LINKED ISSUE #105: meta: codemod situation
## Context

We want to provide serval codemods to help users migrate from older versions of node.js to newer ones. But codemods is currently in a "refactoring" state, and we need to decide how to proceed.

They are marking the `@lasted` npm registry tag to legacy in [their doc](https://docs.codemod.com/deploying-codemods/cli). And they are recommanding to use the `@next` tag instead.

Also codemod team inform me that they will keep `@next` as stable in serval weeks. I don't have any informaiton about a LTS policy for the "legacy" CLI.

Note that this new CLI is a complete rewrite of the CLI in rust it's use QuickJS (so not Node.js) to run codemods. So it's mean when you install it with `npm install codemod@next` you just download a binary.

## Pros

- `@next` CLI include a testing tool, which is very useful to test codemods before deploying them. Without having complex setup. But it's mean that we will depend on codemod to make this initiative work. [documentation](https://github.com/codemod-com/codemod/tree/main/crates/cli#jssg-testing-framework-usage-guide)
- `@next` have a well integration of [`ast-grep`](https://ast-grep.github.io) which is a very powerful & performant tool to manipulate AST. It will allow us to write more complex codemods, and to have a better performance.
- If we write our codemod with `@lasted` we will need to update (small rewrite) them when `@next` will became stable.

### Cons

- ~`@next` CLI use a fresh new codemod registry, and https://codemod.com/registry does not allow to search for codemods on this new registry. So it's will impact our users.~ **See** https://app.codemod.com/registry?q=@nodejs
- User will have to run it with `npx codemod@next` instead of `npx codemod`, which is not a big deal, but it's not the same experience. _I will never recommend a global install of any CLI._
- `@next` can be unstable since there aren't a lot of users yet. But the codemod team is very responsive and they are working on it.
- Previous codemod registry was open source, but the new one is not. So if there are issues with the registry, we will have to wait for the codemod team to fix it.

## Decision

- Should we use `lasted` like in [Correct TypeScript Specifiers](https://github.com/nodejs/userland-migrations/tree/main/recipes/correct-ts-specifiers) codemod?
- Should we use `next` like in [#89](https://github.com/nodejs/userland-migrations/pull/89)
- Do nothing just open issue about which codemod we want to write and wait for the codemod team to stabilize the `@next` CLI?

> do not use our current `@latest` tag for now. just use `@next` for now. use our new workflow engine. publish to our new registry. and thanks so much for your patience over the new few weeks. we hope that you like the end results.
> @alexbit-codemod - co-founder of codemod, message sent on codemod slack  

## Personal goal

I want to ship `node-vX-to-vY` codemods for the next LTS release. And I'm in holliday until september, so i have a lot of time to work on this initiative. But we need to decide how do we want to write our codemods.

## Current state of adoption 

- [x] #108
  - [x] which is depending on https://github.com/nodejs/admin/issues/989
- [x] https://github.com/codemod-com/codemod/pull/1639 which didn't allow use to use our cool utility 
We can bypass this one by bundling on our side (which I'm not fan of that because it's increase the amount of thing to maintain) 

## Thanks

Thanks to all members of @nodejs/userland-migrations
And thanks to the codemod team for their work and their support

aashvgit:


aashvgit:
actually it is for #133

AugustinMauroy:
Hey @aashvgit thanks for giving time on this but we don't use `jscodeshift`. We use [codemod platform](https://codemod.com) though [their new cli](https://codemod.com/blog/new-codemod-cli) based on [ast-grep](https://ast-grep.github.io).
So you should have something that look like this pr https://github.com/nodejs/userland-migrations/pull/89

Also if you need help you can join the [openjs slack](https://slack-invite.openjsf.org/) or [nodejs discord](https://nodejs.org/discord) where I can drive you

PR #135, opened 2025-07-27
Feat(`fs truncate fd depreciation`): introduce

### LINKED ISSUE #122: feat: handle `fs.truncate()` using a file descriptor depreciation
## Description

Since `fs.truncate()` and `fs.truncateSync()` with file descriptors are deprecated (DEP0081) and have reached End-of-Life status, we should provide a codemod to replace them.

- The codemod should replace all instances of `fs.truncate(fd, ...)` with `fs.ftruncate(fd, ...)` when the first argument is a file descriptor.
- The codemod should replace all instances of `fs.truncateSync(fd, ...)` with `fs.ftruncateSync(fd, ...)` when the first argument is a file descriptor.
- The codemod should update import/require statements to include `ftruncate` and `ftruncateSync` when needed.
- The codemod should NOT replace `fs.truncate(path, ...)` calls where the first argument is a file path string.

## Additional Information

Note that `fs.truncate()` and `fs.truncateSync()` with file descriptors are deprecated in favor of `fs.ftruncate()` and `fs.ftruncateSync()`. The behavior is identical, but the new functions have clearer naming to indicate they work with file descriptors. Using `fs.truncate()` with file paths (strings) remains valid and supported.

## Detection Strategy

The codemod should identify file descriptor usage by analyzing the context:

- If the first argument to truncate/truncateSync is a variable that was assigned from fs.open/fs.openSync calls
- If the first argument is a numeric literal or variable with numeric type hints
- If the call is within a callback from fs.open() where the second parameter is typically the file descriptor

### Case 1: Destructured import with callback version

**Before:**
```js
const { truncate, open, close } = require('node:fs');

open('file.txt', 'w', (err, fd) => {
  if (err) throw err;
  truncate(fd, 10, (err) => {
    if (err) throw err;
    close(fd, () => {});
  });
});
```

**After:**
```js
const { ftruncate, open, close } = require('node:fs');

open('file.txt', 'w', (err, fd) => {
  if (err) throw err;
  ftruncate(fd, 10, (err) => {
    if (err) throw err;
    close(fd, () => {});
  });
});
```

### Case 2: Namespace import with sync version

**Before:**
```js
const fs = require('node:fs');

const fd = fs.openSync('file.txt', 'w');
try {
  fs.truncateSync(fd, 10);
} finally {
  fs.closeSync(fd);
}
```

**After:**
```js
const fs = require('node:fs');

const fd = fs.openSync('file.txt', 'w');
try {
  fs.ftruncateSync(fd, 10);
} finally {
  fs.closeSync(fd);
}
```

### Case 3: Mixed usage (file descriptor and file path)

**Before:**
```js
const { truncate, truncateSync, open, openSync, close, closeSync } = require('node:fs');

// This should be replaced (file descriptor)
const fd = openSync('file.txt', 'w');
truncateSync(fd, 10);
closeSync(fd);

// This should NOT be replaced (file path)
truncate('other.txt', 5, (err) => {
  if (err) throw err;
});
```

**After:**
```js
const { truncate, ftruncateSync, open, openSync, close, closeSync } = require('node:fs');

// This should be replaced (file descriptor)
const fd = openSync('file.txt', 'w');
ftruncateSync(fd, 10);
closeSync(fd);

// This should NOT be replaced (file path)
truncate('other.txt', 5, (err) => {
  if (err) throw err;
});
```

### Case 4: ESM import

**Before:**
```js
import { truncate, open, close } from 'node:fs';

open('file.txt', 'w', (err, fd) => {
  if (err) throw err;
  truncate(fd, 10, (err) => {
    if (err) throw err;
    close(fd, () => {});
  });
});
```

**After:**
```js
import { ftruncate, open, close } from 'node:fs';

open('file.txt', 'w', (err, fd) => {
  if (err) throw err;
  ftruncate(fd, 10, (err) => {
    if (err) throw err;
    close(fd, () => {});
  });
});
```

## REFS

[DEP0081: `fs.truncate()` using a file descriptor](https://nodejs.org/docs/latest/api/deprecations.html#DEP0081)

AugustinMauroy:
## Related issue

close #122


avivkeller (dismissed):
It's "Deprecation", not "Depreciation"

AugustinMauroy:
> It's "Deprecation", not "Depreciation"

🤦🏼

brunocroh (approved):

recipes/fs-truncate-fd-deprecation/src/workflow.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  // Check if it's assigned from fs.openSync or openSync
+  if (isAssignedFromOpenSync(param, rootNode)) return true;
+
+  // Check if it's used inside a callback context from fs.open
+  if (isInCallbackContext(param, rootNode)) return true;
+
+  // For other cases, be conservative - don't transform unless we're sure
+  return false;
+}
```
avivkeller, line 196: nit: this can be simplified slightly by merging one or two if conditions

avivkeller (approved):

recipes/fs-truncate-fd-deprecation/src/workflow.ts
AugustinMauroy: Yeah you had right but written like that it's verbose and make easy to read/understand

avivkeller: At least merge the last two using "OR", you can still add comments, it'll just run a bit faster since there are less computations

JakobJingleheimer (approved):
🙌

JakobJingleheimer (approved):

PR #136, opened 2025-07-27
feat(process-main-module): introduce

### LINKED ISSUE #120: feat: handle `process.mainModule` depreciation
## Description

Since `process.mainModule` is deprecated in Node.js v14.0.0, we should provide a codemod to replace it.

- The codemod should modify any usage of this api.
- If `const { mainModule } = require('node:process')` is used need to check if used.

## Examples

**Before:**

```js
if (process.mainModule === 'mod.js') {
  // cli thing
} else {
  // module thing
}
```

**After:**

```js
if (require.main === 'mod.js') {
  // cli thing
} else {
  // module thing
}
```

## REFS

- [`DEP0138: process.mainModule`](https://nodejs.org/api/deprecations.html#DEP0138)

brunocroh:
Closes #120 

I'm currently ignoring a TypeScript validation because the Range type from `@codemod.com/jssg-types/main` is incorrect, which affects the types in utils/removeLine.

### DIFF
--- a/recipes/process-main-module/src/workflow.ts
+++ b/recipes/process-main-module/src/workflow.ts
@@ -1,0 +1,80 @@
+import type { Edit, SgRoot, Range } from "@codemod.com/jssg-types/main";
+import { getNodeRequireCalls } from "@nodejs/codemod-utils/ast-grep/require-call";
+import { removeLines } from "@nodejs/codemod-utils/ast-grep/remove-lines";
+
+/**
+ * Transforms the usage of `process.mainModule` to use the `require.main`. *
+ *
+ * Handles:
+ * 1. Find all destructuring require statements from 'node:process' module that import 'mainModule'
+ *
+ * 2. Handle the destructuring import:
+ *    - If 'mainModule' is the only imported property → Remove the entire require statement
+ *    - If other properties are also imported → Remove only 'mainModule' from the destructuring
+ *
+ * 3. Replace all code references:
+ *    - Change 'mainModule' → 'require.main'
+ *    - Change 'process.mainModule' → 'require.main'
+ */
+export default function transform(root: SgRoot): string | null {
+	const rootNode = root.root();
+	const edits: Edit[] = [];
+	const linesToRemove: Range[] = [];
+
+	// @ts-ignore - ast-grep types are not fully compatible with JSSG types
+	const requireDeclarations = getNodeRequireCalls(root, "process");
+
+	for (const declarationNode of requireDeclarations) {
+		// Step 1: Get all requires from module nodule:process that is destructuring mainModule:
+		if (declarationNode.text().includes("mainModule")) {
+			const objectPattern = declarationNode.find({
+				rule: {
+					kind: "object_pattern",
+				},
+			});
+
+			if (!objectPattern) continue;
+
+			// Step2: Handle the destructuring import:
+			const declarations = declarationNode.findAll({
+				rule: {
+					kind: "shorthand_property_identifier_pattern",
+				},
+			});
+
+			if (declarations.length === 1) {
+				const nodes = rootNode.findAll({
+					rule: {
+						pattern: "mainModule",
+					},
+				});
+				linesToRemove.push(declarationNode.range());
+				nodes.forEach((node) => {
+					edits.push(node.replace("require.main"));
+				});
+			}
+
+			if (declarations.length > 1) {
+				const restDeclarations = declarations
+					.map((d) => d.text())
+					.filter((d) => d !== "mainModule");
+				edits.push(objectPattern.replace(`{ ${restDeclarations.join(", ")} }`));
+			}
+		}
+	}
+
+	// Step 3: Replace all code references:
+	const nodes = rootNode.findAll({
+		rule: {
+			pattern: "process.mainModule",
+		},
+	});
+
+	nodes.forEach((node) => {
+		edits.push(node.replace("require.main"));
+	});
+
+	let sourceCode = rootNode.commitEdits(edits);
+	sourceCode = removeLines(sourceCode, linesToRemove);
+	return sourceCode;
+}
--- a/recipes/process-main-module/src/workflow.ts
+++ b/recipes/process-main-module/src/workflow.ts
@@ -22,6 +22,11 @@ export default function transform(root: SgRoot): string | null {
 	const rootNode = root.root();
 	const edits: Edit[] = [];
 	const linesToRemove: Range[] = [];
+	const patternsToReplace = [
+		{
+			pattern: "process.mainModule",
+		},
+	];
 
 	// @ts-ignore - ast-grep types are not fully compatible with JSSG types
 	const requireDeclarations = getNodeRequireCalls(root, "process");
@@ -61,6 +66,10 @@ export default function transform(root: SgRoot): string | null {
 					.map((d) => d.text())
 					.filter((d) => d !== "mainModule");
 
+				patternsToReplace.push({
+					pattern: "mainModule",
+				});
+
 				edits.push(objectPattern.replace(`{ ${restDeclarations.join(", ")} }`));
 			}
 		}
@@ -69,14 +78,7 @@ export default function transform(root: SgRoot): string | null {
 	// Step 3: Replace all code references:
 	const nodes = rootNode.findAll({
 		rule: {
-			any: [
-				{
-					pattern: "process.mainModule",
-				},
-				{
-					pattern: "mainModule",
-				},
-			],
+			any: patternsToReplace,
 		},
 	});
 
--- a/recipes/process-main-module/src/workflow.ts
+++ b/recipes/process-main-module/src/workflow.ts
@@ -62,26 +62,21 @@ export default function transform(root: SgRoot): string | null {
 				},
 			});
 
-			if (declarations.length === 1) {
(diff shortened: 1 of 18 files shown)


recipes/process-main-module/.gitignore
AugustinMauroy, line 1: I don't think that's necessary because the one at root normally covers that. If it doesn't, merge the two together 😁

recipes/process-main-module/workflow.yaml
```diff
@@ -0,0 +1,24 @@
+version: "1"
+nodes:
```
AugustinMauroy, line 2:
```suggestion

nodes:
```

let's add little bit of space to make it prettier

utils/src/ast-grep/remove-lines.ts
```diff
@@ -0,0 +1,20 @@
+import type { Range } from "@ast-grep/napi";
+
+export function removeLines(sourceCode: string, ranges: Range[]) {
+	const lines = sourceCode.split("\n");
+
+	let removeCounter = 0;
+	for (const range of ranges) {
+		// @ts-ignore - @ast-grep/napi is returning start.row, but the type says the field need to be
```
AugustinMauroy, line 8: @mohebifar I know you worked on codemod and ast-grep any input here ?

recipes/process-main-module/src/workflow.ts
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	// Step 3: Replace all code references:
+	const nodes = rootNode.findAll({
+		rule: {
+			pattern: "process.mainModule",
+		},
+	});
+
+	nodes.forEach((node) => {
```
AugustinMauroy, line 73: you should use 
```js
for (const node of nodes) {
}
```

recipes/process-main-module/src/workflow.ts
```diff
@@ -0,0 +1,80 @@
+import type { Edit, SgRoot, Range } from "@codemod.com/jssg-types/main";
+import { getNodeRequireCalls } from "@nodejs/codemod-utils/ast-grep/require-call";
+import { removeLines } from "@nodejs/codemod-utils/ast-grep/remove-lines";
+
+/**
+ * Transforms the usage of `process.mainModule` to use the `require.main`. *
+ *
```
AugustinMauroy, line 7: add note that this api is CJS only so it's why we only care about `require` importing

AugustinMauroy (commented):
WOW that's great !

recipes/process-main-module/.gitignore
AugustinMauroy: By the time I finish the review, you'll have taken it down, okay

utils/src/ast-grep/remove-lines.ts
brunocroh: https://github.com/codemod-com/codemod/pull/1655

I opened a PR on their repo about it

recipes/process-main-module/.gitignore
brunocroh: I notice it right after push, sorry 😅

recipes/process-main-module/src/workflow.ts
brunocroh: Done, could you give me details about this change request? Is it performance related, code style or anything else? I think I can learn something here, thank you.

utils/src/ast-grep/remove-lines.ts
AugustinMauroy: I mean adding the pr link the codemod one in the code comment aside of `@ts-ignore`

brunocroh: I got it know, I will do it so.

AugustinMauroy:
Okay, to give you some feedback, we're waiting for a second review of your PR before merging it. Then we'll release the codemod.

recipes/process-main-module/src/workflow.ts
brunocroh: What do you think about if the comment become:

 ```
 Transforms `process.mainModule` usage to `require.main`. Handles direct global access
 with `process.mainModule` or `require("node:process")` imports, as `mainModule` is CJS-only.
 ```
 
 And I think the README.md need to be updated from below one, to it above as well, what do you think?
 ```
 This recipe transforms the usage of `process.mainModule` to use the `require.main`.

 ```

recipes/process-main-module/tests/expected/importing-multiple-values.js
```diff
@@ -0,0 +1,9 @@
+const { env, cwd } = require("node:process"); // mainModule as first
+const { env, cwd } = require("node:process"); // mainModule as last
+const { env, cwd } = require("node:process"); // mainModule at middle
+
+if (mainModule === module) {
```
avivkeller, line 5:
```suggestion
if (require.main === module) {
```

Right?

AugustinMauroy: yeah right, good catch aviv

recipes/process-main-module/src/workflow.ts
```diff
@@ -68,7 +69,14 @@ export default function transform(root: SgRoot): string | null {
...
 	const nodes = rootNode.findAll({
 		rule: {
-			pattern: "process.mainModule",
+			any: [
+				{
+					pattern: "process.mainModule",
+				},
+				{
+					pattern: "mainModule",
+				},
```
avivkeller, line 78: We should only match this if it was found in an import / destructure

avivkeller (commented):
Can you add a test for:

````js
const { mainModule } = process

if (mainModule) {

}
````

File: recipes/process-main-module/src/workflow.ts (hunk 6 above)
Comment (AugustinMauroy) [reply recipes/process-main-module/src/workflow.ts:78]: yeah agreed and maybe have two query on to gets `mainModule` and another for the requiring

Comment (AugustinMauroy) [reply recipes/process-main-module/src/workflow.ts:78]: what do you mean ? here it's only for usage part

Comment (brunocroh) [reply recipes/process-main-module/src/workflow.ts]: I created patternsToReplace to handle it

**Comment by brunocroh** [comment]
> Can you add a test for:
> 
> ```js
> const { mainModule } = process
> 
> if (mainModule) {
> 
> }
> ```

Good catch, done!

**Comment by brunocroh** [comment]
FYI: [fix(codemod-sandbox): Replace row with line in JsPosition to match the types from @ast-grep/napi #1655](https://github.com/codemod-com/codemod/pull/1655) have been merged and will be available in version `1.0.0-rc.19`. With that release, we’ll be able to remove the `@ts-ignore` previously required here. 

Let me know if there's anything else I should update!

**Comment by AugustinMauroy** [comment]
19 is released so I think you can remove your comment

**Comment by brunocroh** [comment]
> 19 is released so I think you can remove your comment

done

File: recipes/process-main-module/workflow.yaml [hunk 7]
```diff
@@ -0,0 +1,25 @@
...
+          base_path: .
+          include:
+            - "**/*.js"
+            - "**/*.jsx"
+            - "**/*.mjs"
+            - "**/*.cjs"
+            - "**/*.cts"
+            - "**/*.mts"
+            - "**/*.ts"
+            - "**/*.tsx"
```
JakobJingleheimer, line 22: nit: alphabetical order (for grokking)

```suggestion
            - "**/*.cjs"
            - "**/*.js"
            - "**/*.jsx"
            - "**/*.mjs"
            - "**/*.cts"
            - "**/*.mts"
            - "**/*.ts"
            - "**/*.tsx"
```

JakobJingleheimer (commented):
Thanks @brunocroh! Nice robust test cases 😁

Mostly small things and a question. Looking good!

recipes/process-main-module/src/workflow.ts
```diff
@@ -0,0 +1,105 @@
...
+				const restDeclarations = declarations
+					.map((d) => d.text())
+					.filter((d) => d !== "mainModule");
+
+				patternsToReplace.push({
+					pattern: "mainModule",
+				});
+
+				edits.push(objectPattern.replace(`{ ${restDeclarations.join(", ")} }`));
+			}
```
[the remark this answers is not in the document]
brunocroh: The difference is that when only one element exists in the destructured object, we remove the entire import line.

```js
const { mainModule } = require("node:process");
const { env, mainModule } = require("node:process");
```

In the example above, the first line will be removed, and the second one will become:

```js
const { env } = require("node:process");
```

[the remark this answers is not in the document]
JakobJingleheimer: OH! Good thinking. Could you please add a brief code-comment above the `1` case explaining that? (Something simple like `When there's only 1 item imported, remove to whole thing to avoid an empty import`)

recipes/process-main-module/src/workflow.ts
JakobJingleheimer, line 1: Nice. Easier to follow now :)

JakobJingleheimer (approved):
LGTM 🙌

JakobJingleheimer:
> Yes, of course! If I'm doing anything wrong or if there's something I could improve, please feel free to let me know. I'm still learning how to contribute, and I'm really enjoying it so far.

No worries! This repo is different from node core (where you _do_ have to force push some times 🤪); hopefully this is easier.

Thanks for the contribution, and glad to hear you're enjoying it!

PR #139, opened 2025-07-27
chore: update biome schema path

brunocroh:
To prevent Biome from warning about an incorrect schema version, update the Biome configuration to use the locally installed schema.

Screenshot 2025-07-27 at 12 47 02


AugustinMauroy (approved):
LGMT clean

PR #140, opened 2025-07-27
setup(ci-publish): use `@next` codemod cli

### LINKED ISSUE #108: CI: update to support `@next` codemod cli
This file need to be updated to use `@next` since in #105 we will use this new cli

https://github.com/nodejs/userland-migrations/blob/main/.github/workflows/codemod_publish.yml
https://github.com/codemod-com/codemod/blob/main/crates/cli/README.md

AugustinMauroy:
## Description 

introduce new ci for releasing codemod based on git tag

Also not that `correct-ts-specifiers` will not be anymore publish using GHA since it's use legacy codemod.

## Related issue 

Close #108

### DIFF
--- a/.github/workflows/codemod_publish.yml
+++ b/.github/workflows/codemod_publish.yml
@@ -43,7 +43,7 @@ jobs:
           echo "CODEMOD_NAME=$CODEMOD_NAME" >> $GITHUB_OUTPUT
           echo "CODEMOD_PATH=recipes/$CODEMOD_NAME" >> $GITHUB_OUTPUT
 
-          echo "Publishing: $CODEMOD_NAME (version: $VERSION) from recipes/$CODEMOD_NAME"
+          echo "Publishing version $VERSION of $CODEMOD_NAME from recipes/$CODEMOD_NAME"
 
       - name: Validate codemod directory exists
         run: |
--- a/.github/workflows/codemod_publish.yml
+++ b/.github/workflows/codemod_publish.yml
@@ -64,14 +64,15 @@ jobs:
       - name: Install Codemod CLI
         run: npm install -g codemod@next
 
+      # Run test first so if it's failing, we don't waste time to login and rest of thing
+      - name: Run test
+        working-directory: ${{ steps.extract.outputs.CODEMOD_PATH }}
+        run: node --run test
+
       - name: Login to Codemod registry
         env:
           CODEMOD_TOKEN: ${{ secrets.CODEMOD_TOKEN }}
         run: codemod login --token "$CODEMOD_TOKEN"
-
-      - name: Run test
-        working-directory: ${{ steps.extract.outputs.CODEMOD_PATH }}
-        run: node --run test
 
       - name: Publish codemod (dry run)
         working-directory: ${{ steps.extract.outputs.CODEMOD_PATH }}
--- a/.github/workflows/codemod_publish.yml
+++ b/.github/workflows/codemod_publish.yml
@@ -77,13 +77,13 @@ jobs:
           if [[ ! -d "$CODEMOD_PATH" ]]; then
             echo "❌ Codemod directory not found: $CODEMOD_PATH"
             echo "Available directories in recipes/:"
-            ls -la recipes/ || echo "No recipes directory found"
+            ls -lah recipes/ || echo "No recipes directory found"
             exit 1
           fi
 
-          echo "✅ Found codemod directory: $CODEMOD_PATH"
+          echo "✓ Found codemod directory: $CODEMOD_PATH"
           echo "Directory contents:"
-          ls -la "$CODEMOD_PATH"
+          ls -lah "$CODEMOD_PATH"
 
       - name: Setup Node.js environment
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
--- a/.github/workflows/codemod_publish.yml
+++ b/.github/workflows/codemod_publish.yml
@@ -92,11 +92,11 @@ jobs:
           cache: npm
           cache-dependency-path: package-lock.json
 
+      # We don't use dev dependencies
+      # But we need npm to put local workspace in `node_modules`
+      # so codemod can bundle workspaces in the codemod tarball correctly
       - name: Install project dependencies
         run: npm ci
-
-      - name: Install Codemod CLI
-        run: npm install -g codemod@next
 
       # Run test before login to not waste time if it fails
       - name: Run codemod Tests
@@ -106,11 +106,11 @@ jobs:
       - name: Authenticate with Codemod registry
         env:
           CODEMOD_TOKEN: ${{ secrets.CODEMOD_TOKEN }}
-        run: codemod login --token "$CODEMOD_TOKEN"
+        run: npx codemod@next login --token "$CODEMOD_TOKEN"
 
       - name: Publish codemod
         working-directory: ${{ steps.parse-tag.outputs.codemod-path }}
-        run: codemod publish
+        run: npx codemod@next publish
 
       - name: Create release summary
         env:
--- a/.github/workflows/codemod_publish.yml
+++ b/.github/workflows/codemod_publish.yml
@@ -1,107 +1,82 @@
-name: Codemod publish
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+# For more information see: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow
+
+name: Publish Codemod
 
 on:
-  workflow_dispatch:
   push:
-    paths:
-      - "recipes/**"
-    branches:
-      - main
-
-permissions:
-  contents: read
+    tags:
+      - 'v*@*'  # Match tags like v1.0.0@my-codemod
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag to publish (e.g., v1.0.0@my-codemod)'
+        required: true
+        type: string
 
 jobs:
-  paths-filter:
-    permissions:
-      contents: read
-      pull-requests: read
-    name: Check for .codemodrc.json files changes
-    runs-on: ubuntu-latest
-    outputs:
-      codemods: ${{ steps.filter.outputs.codemods }}
-      codemods_files: ${{ steps.filter.outputs.codemods_files }}
-    steps:
-      - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49
-        with:
-          egress-policy: audit
-
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
-      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36
-        id: filter
-        name: Filter codemods
-        with:
-          list-files: json
-          filters: |
-            codemods:
-              - '**/.codemodrc.json'
-
-  prepare-matrix:
-    name: Prepare matrix
-    runs-on: ubuntu-latest
-    needs: paths-filter
-    if: always() && needs.paths-filter.outputs.codemods == 'true'
-    outputs:
-      codemod_files: ${{ steps.set-matrix.outputs.codemod_files }}
-    steps:
-      - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49
-        with:
-          egress-policy: audit
-
-      - name: Checkout Code
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
-        with:
-          fetch-depth: 0
-
-      - name: Set matrix for codemods
-        id: set-matrix
-        run: |
-          FILES_JSON=$(echo '${{ needs.paths-filter.outputs.codemods_files }}' | jq -c '{include: map({file: .})}')
-          echo "codemod_files=$FILES_JSON" >> "$GITHUB_OUTPUT"
-          echo "Matrix JSON: $FILES_JSON"
-
   publish:
-    name: Publish ${{ matrix.file }}
     runs-on: ubuntu-latest
-    needs: prepare-matrix
-    strategy:
-      fail-fast: false
-      matrix: ${{fromJson(needs.prepare-matrix.outputs.codemod_files)}}
-    env:
-      CODEMOD_API_KEY: ${{ secrets.CODEMOD_API_KEY }}
+
     steps:
-      - name: Harden the runner (Audit all outbound calls)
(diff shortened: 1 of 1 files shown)


AugustinMauroy:
I need explicit Jacob approval since he is responsible of repo secret.
@JakobJingleheimer ping me when it's done I'll delete my personnal api token (used for manual publish)

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
-      codemods: ${{ steps.filter.outputs.codemods }}
-      codemods_files: ${{ steps.filter.outputs.codemods_files }}
+
     steps:
-      - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49
-        with:
-          egress-policy: audit
+      - name: Checkout code
+        uses: actions/checkout@v4
```
JakobJingleheimer, line 24: For an abundance of caution, let's use the release hash (dependabot will handle updating it after this is committed with a hash).

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
-          filters: |
-            codemods:
-              - '**/.codemodrc.json'
+      - name: Extract version and codemod name
+        id: extract
+        run: |
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
+            TAG="${{ github.event.inputs.tag }}"
+          else
+            TAG="${GITHUB_REF#refs/tags/}"
```
JakobJingleheimer, line 32: It's not possible for a non-member to create a tag, right? (I'm thinking this could be a potential attack vector)

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
-          egress-policy: audit
+          # Extract version (everything before @)
+          VERSION="${TAG%@*}"
+          VERSION="${VERSION#v}"  # Remove 'v' prefix
 
-      - name: Checkout Code
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
-        with:
-          fetch-depth: 0
+          # Extract codemod name (everything after @)
```
JakobJingleheimer, line 39: Won't it need to strip the `nodejs` too?

Ex below where it says

```
"Publishing: $CODEMOD_NAME (version: $VERSION) from recipes/$CODEMOD_NAME"
```

That will print something like
```
Publishing: nodejs/foo-to-bar (version: 1.0.0) from recipes/nodejs/foo-to-bar`).
```

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
-      fail-fast: false
-      matrix: ${{fromJson(needs.prepare-matrix.outputs.codemod_files)}}
-    env:
-      CODEMOD_API_KEY: ${{ secrets.CODEMOD_API_KEY }}
-    steps:
-      - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49
-        with:
-          egress-policy: audit
+          echo "Publishing: $CODEMOD_NAME (version: $VERSION) from recipes/$CODEMOD_NAME"
```
JakobJingleheimer, line 46:
```suggestion
          echo "Publishing version $VERSION of $CODEMOD_NAME from recipes/$CODEMOD_NAME"
```

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
+      - name: Validate codemod directory exists
+        run: |
+          if [ ! -d "${{ steps.extract.outputs.CODEMOD_PATH }}" ]; then
+            echo "Error: Directory ${{ steps.extract.outputs.CODEMOD_PATH }} does not exist"
+            exit 1
+          fi
 
-      - uses: actions/setup-node@cdca7365b2dadb8aad0a33bc7601856ffabcc48e
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
```
JakobJingleheimer, line 56: Same (let's use the hash)

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
 
-      - name: Test codemod
-        run: |
-          DIR=$(dirname "${{ matrix.file }}")
-          echo "Testing codemod in: $DIR"
-          cd "$DIR"
-          npm install
-          npm run --if-present test
+      - name: Install Codemod CLI
+        run: npm install -g codemod@next
```
JakobJingleheimer, line 65: I think we don't need to install it globally?

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
+
+      - name: Login to Codemod registry
+        env:
+          CODEMOD_TOKEN: ${{ secrets.CODEMOD_TOKEN }}
+        run: codemod login --token "$CODEMOD_TOKEN"
+
+      - name: Run test
```
JakobJingleheimer, line 72: I think running tests should come before logging into codemode: if the test run fails, there's no need to log in (so it's just wasted time).

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
+        working-directory: ${{ steps.extract.outputs.CODEMOD_PATH }}
+        run: node --run test
+
+      - name: Publish codemod (dry run)
+        working-directory: ${{ steps.extract.outputs.CODEMOD_PATH }}
+        run: codemod publish --dry-run
```
JakobJingleheimer, line 78: Why a dry-run right before the full?

avivkeller (dismissed):

avivkeller:
Blocking, since I have some security concerns

avivkeller:
@nodejs/userland-migrations Can someone with the appropriate access disallow:
- Merging without approvals
- Merging when changes have been requested
- Merging when checks are failing

.github/workflows/codemod_publish.yml
avivkeller: Correct. Only users with `write` access to this repository can add tags.

JakobJingleheimer:
> I need explicit Jacob approval since he is responsible of repo secret. @JakobJingleheimer ping me when it's done I'll delete my personnal api token (used for manual publish)

I already added the nodejs token to the repo secret during our call on Saturday. Is it not working?

avivkeller:
I think @AugustinMauroy has the wrong secret name. It's `CODEMOD_PUBLICATION_KEY`

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
-          echo "Testing codemod in: $DIR"
-          cd "$DIR"
-          npm install
-          npm run --if-present test
+      - name: Install Codemod CLI
+        run: npm install -g codemod@next
+
+      - name: Login to Codemod registry
+        env:
+          CODEMOD_TOKEN: ${{ secrets.CODEMOD_TOKEN }}
```
avivkeller, line 69:
```suggestion
          CODEMOD_TOKEN: ${{ secrets.CODEMOD_PUBLICATION_KEY }}
```

.github/workflows/codemod_publish.yml
AugustinMauroy: only people with write role can create a tag.

.github/workflows/codemod_publish.yml
JakobJingleheimer: Ahh, yes indeed :)

.github/workflows/codemod_publish.yml
AugustinMauroy: it's allow us to be sure we use the same version by workflow run.

JakobJingleheimer:
@avivkeller I think the protections are in place now (I thought I had already done).

btw @AugustinMauroy thanks for tackling this!

.github/workflows/codemod_publish.yml
avivkeller: @AugustinMauroy Anyone with write access can define the name, fwiw

.github/workflows/codemod_publish.yml
JakobJingleheimer: I think we can do both: first fix the things we know about, and then check for things we don't know about. Otherwise, there is a however brief window of vulnerability to things we already could have fixed.

.github/workflows/codemod_publish.yml
JakobJingleheimer: I think it doesn't actually do that?

I think if you use `npx codemod@next …` that will ensure it uses the same thing.

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
-        name: Filter codemods
-        with:
-          list-files: json
-          filters: |
-            codemods:
-              - '**/.codemodrc.json'
+      - name: Extract version and codemod name
+        id: extract
+        run: |
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
```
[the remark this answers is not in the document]
JakobJingleheimer: They're supposed to go into an ENV var first, right? (and then read from the env var?)

.github/workflows/codemod_publish.yml
AugustinMauroy: the comment may be not 100% right but it's node codemod name it's codemod dirname such as `correct-ts-specifiers`

tag will look like that `v1.0.1@correct-ts-specifiers` and `CODEMOD_NAME` will be `correct-ts-specifiers`

.github/workflows/codemod_publish.yml
AugustinMauroy: because the cli isn't 100% stable i want to run one time to unsere we will not publish an ivalid tarball because cli don't cut the process before sending it 

(BTW CLI din't have any test)

.github/workflows/codemod_publish.yml
AugustinMauroy: good point

.github/workflows/codemod_publish.yml
AugustinMauroy: use `CODEMOD_PUBLICATION_KEY`

.github/workflows/codemod_publish.yml
```diff
@@ -64,15 +64,16 @@ jobs:
       - name: Install Codemod CLI
         run: npm install -g codemod@next
 
-      - name: Login to Codemod registry
-        env:
-          CODEMOD_TOKEN: ${{ secrets.CODEMOD_TOKEN }}
-        run: codemod login --token "$CODEMOD_TOKEN"
-
+      # Run test first so if it's failing, we don't waste time to login and rest of thing
```
JakobJingleheimer, line 67: ❤️

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,82 @@
...
-      pull-requests: read
-    name: Check for .codemodrc.json files changes
+  publish:
     runs-on: ubuntu-latest
-    outputs:
-      codemods: ${{ steps.filter.outputs.codemods }}
-      codemods_files: ${{ steps.filter.outputs.codemods_files }}
+
     steps:
-      - name: Harden the runner (Audit all outbound calls)
```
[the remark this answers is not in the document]
AugustinMauroy: Intesting

.github/workflows/codemod_publish.yml
AugustinMauroy: oh that found there ... I'll go to bed a little wiser.

AugustinMauroy:
> @AugustinMauroy I force-pushed by accident, hopefully I didn't break anything. All my security concerns are now resolved

Let me check that out, but it's great because I have to admit that GHAs are a necessary evil but at the same time a nightmare.

.github/workflows/codemod_publish.yml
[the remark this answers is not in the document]
AugustinMauroy: let's not use for now to keep this repo away for that

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,132 @@
...
+          echo "codemod-path=$CODEMOD_PATH" >> $GITHUB_OUTPUT
+
+      - name: Verify codemod directory
+        env:
+          CODEMOD_PATH: ${{ steps.parse-tag.outputs.codemod-path }}
+        run: |
+          if [[ ! -d "$CODEMOD_PATH" ]]; then
+            echo "❌ Codemod directory not found: $CODEMOD_PATH"
+            echo "Available directories in recipes/:"
+            ls -la recipes/ || echo "No recipes directory found"
```
JakobJingleheimer, line 80: very nit: add `h` ("human" friendly)

```suggestion
            ls -lah recipes/ || echo "No recipes directory found"
```

AugustinMauroy:
I think I had resolved whole of your concern

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,132 @@
...
+            exit 1
+          fi
 
-      - uses: actions/setup-node@cdca7365b2dadb8aad0a33bc7601856ffabcc48e
+          echo "✅ Found codemod directory: $CODEMOD_PATH"
```
JakobJingleheimer, line 84: very nit: the green checkmark is "too much success" 😅 this is only a small success

```suggestion
          echo "✓ Found codemod directory: $CODEMOD_PATH"
```

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,132 @@
...
+          echo "Directory contents:"
+          ls -la "$CODEMOD_PATH"
```
JakobJingleheimer, line 86:
```suggestion
          ls -lah "$CODEMOD_PATH"
```

JakobJingleheimer (approved):
LGTM now 🙂

.github/workflows/codemod_publish.yml
AugustinMauroy: OOOHHHH

JakobJingleheimer (approved):

.github/workflows/codemod_publish.yml
```diff
@@ -1,107 +1,132 @@
...
+            echo "❌ Codemod directory not found: $CODEMOD_PATH"
+            echo "Available directories in recipes/:"
+            ls -lah recipes/ || echo "No recipes directory found"
+            exit 1
+          fi
 
-      - uses: actions/setup-node@cdca7365b2dadb8aad0a33bc7601856ffabcc48e
+          echo "✓ Found codemod directory: $CODEMOD_PATH"
+          echo "Directory contents:"
+          ls -lah "$CODEMOD_PATH"
```
avivkeller, line 86:
```suggestion
          tree "$CODEMOD_PATH"
```

AugustinMauroy: tree isn't disponible by default on GHA (need to `apt-get` so for now keep ls)

PR #149, opened 2025-07-29
Feat(`repl builtin modules`): introduce 

### LINKED ISSUE #121: feat: handle `repl.builtinModules` depreciation
## Description

Since `repl.builtinModules` is deprecated in Node.js v24.2.0, we should provide a codemod to replace it.

- The codemod should replace all instances of `repl.builtinModules` with `module.builtinModules`.
- The codemod should also replace `const repl = require('node:repl');` with `const module = require('node:module');` (also ESM import statement) if it exists in the file. And if repl is used for something else, don't replace it.
- if the `node:repl` import is destructured, it should be replaced with `node:module` destructured import. And it's should not replace the whole import if there are other imports from `node:repl` that are not `builtinModules`.

## Additional Information

Note that `repl.builtinModules` is deprecated in favor of `module.builtinModules`. But `module.builtinModules` have more api listed than `repl.builtinModules`, so we need to document that so if the user is using `repl.builtinModules` to get the list of builtin modules, they should be aware that the list will be different.

### Case 1:

**Before:**
```js
const { builtinModules } = require('node:repl');

console.log(builtinModules);
```

**After:**

```js
const { builtinModules } = require('node:module');

console.log(builtinModules);
```

### Case 2:

**Before:**

```js
const repl = require('node:repl');

console.log(repl.builtinModules);
```

**After:**

```js
const module = require('node:module');

console.log(module.builtinModules);
```

### Case 3:

**Before:**

```js
const { builtinModules, foo } = require('node:repl');

console.log(builtinModules);

foo(); // does something else
```

**After:**

```js
const { foo } = require('node:repl');
const { builtinModules } = require('node:module');

console.log(builtinModules);

foo(); // does something else
```

## REFS

- [DEP0191: `repl.builtinModules`](https://nodejs.org/docs/latest-v24.x/api/deprecations.html#dep0191-replbuiltinmodules)

AugustinMauroy:
## Description 

Add new codemod to handle this depreciation 

## Related issue 

Close #121 

## TODO:

- support of https://nodejs.org/api/deprecations.html#DEP0142


recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,438 @@
...
+                        rule: {
+                            pattern: `${varName}.builtinModules`
+                        }
+                    });
+
+                    const builtinLibsExpressions = rootNode.findAll({
+                        rule: {
+                            pattern: `${varName}._builtinLibs`
+                        }
+                    });
```
brunocroh, line 409:
```suggestion
                    const allUsages = rootNode.findAll({
                        rule: {
                        any: [
                          {
                              pattern: `${varName}.builtinModules`
                          }, 
                          {
                              pattern: `${varName}._builtinLibs`
                          }
                    });
```

I think `any` can be used here to simplified it

brunocroh:
I think this recipe can be a great use case for the new utility function when it is ready to use https://github.com/nodejs/userland-migrations/pull/151

recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,321 @@
...
+export default function transform(root: SgRoot<Js>): string | null {
+    const rootNode = root.root();
+    const edits: Edit[] = [];
+
+    // Helper functions
+    const getNewModuleSpecifier = (currentModule: string): string =>
+        currentModule.includes("node:") ? "'node:module'" : "'module'";
+
+    const containsBuiltinProperties = (text: string): boolean =>
+        text.includes("builtinModules") || text.includes("_builtinLibs");
```
avivkeller, line 33: These can go outside of the transform function, to prevent recreating them on any transformation

recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,321 @@
...
+        if (moduleSpecifier) {
+            const currentModule = moduleSpecifier.text();
+            const newModule = getNewModuleSpecifier(currentModule);
+
+            edits.push(moduleSpecifier.replace(newModule));
+        }
+    };
+
+    // Step 1: Handle require statements
+    // @ts-ignore - ast-grep types are not fully compatible with JSSG types
```
avivkeller, line 66: Is there a ref you can link?

recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,321 @@
...
+                const properties = objectPattern.findAll({
+                    rule: { kind: "shorthand_property_identifier_pattern" }
+                });
+                const pairProperties = objectPattern.findAll({
+                    rule: { kind: "pair_pattern" }
+                });
+
+                // Check if only builtin properties are destructured
+                const builtinShorthandCount = properties.filter(p =>
+                    ["builtinModules", "_builtinLibs"].includes(p.text())
```
avivkeller, line 85: Why don't we use the previously defined helper here?

recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,321 @@
...
+                            // Extract alias from pair pattern like "builtinModules: alias" or "_builtinLibs: alias"
+                            const keyNode = prop.find({ rule: { kind: "property_identifier" } });
+                            const valueNode = prop.find({ rule: { kind: "identifier" } });
+
+                            if (
+								keyNode &&
+								valueNode &&
+                                (
+									keyNode.text() === "builtinModules" ||
+									keyNode.text() === "_builtinLibs"
```
avivkeller, line 131: Ditto

avivkeller (commented):
Started a review.

There's a lot of places where you check both properties, those can probably be refactored to common helpers

recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,309 @@
...
+    };
+
+    const updateModuleSpecifier = (statement: SgNode): void => {
+        const moduleSpecifier = statement.find({ rule: { kind: "string" } });
+
+        if (moduleSpecifier) {
+            // Always use 'node:module'
+            const newModule = "'node:module'";
+            edits.push(moduleSpecifier.replace(newModule));
+        }
```
brunocroh, line 56: If you use string_fragment, you get only the text inside single or double quotes, without including the quotes themselves. This way, you can replace the string’s content without changing the user’s preference for single or double quotes.

Nit: The newModule variable can be removed for simplicity.
```suggestion
        const moduleSpecifier = statement.find({ rule: { kind: "string_fragment" } });

        if (moduleSpecifier) {
            // Always use 'node:module'
            edits.push(moduleSpecifier.replace('node:module'));
        }
```

recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,309 @@
...
+                        }
+                    }
+
+                    if (propertiesToKeep.length > 0) {
+                        const variableDeclaration = statement.parent();
+                        const moduleSpecifier = statement.find({ rule: { kind: "string" } });
+
+                        if (variableDeclaration && moduleSpecifier) {
+                            const currentModule = moduleSpecifier.text();
+                            const newModule = "'node:module'";
```
brunocroh, line 137: In a previous comment, I said that const can be removed as not used in other place, looking now, it can be created in a higher scope (begin of transform function) to be reused through the code.

```suggestion
                            const newModule = "'node:module'";
```

AugustinMauroy:
missing dynamic import

recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,404 @@
...
+                                        regex: "(node:)?repl$"
+                                    }
+                                }
+                            }
+                        }
+                    ]
+                }
+            }
+        }
+    });
```
JakobJingleheimer: Woof 🥵 maybe extract this? when reading this code, I don't really care about this part: the const name sufficiently tells me what to expect (and if I really wanna see it, there's always <kbd>cmd</kbd> + click).

recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,404 @@
...
+        const stringFragment = variableDeclarator.find({
+            rule: {
+                kind: "string_fragment",
+                regex: "(node:)?repl$"
+            }
+        });
+
+        if (stringFragment) {
+            edits.push(stringFragment.replace("node:module"));
+        }
```
JakobJingleheimer: nit

```suggestion
        if (stringFragment) edits.push(stringFragment.replace("node:module"));
```

JakobJingleheimer (approved):
Thanks for this!

Some code quality things, but nothing blocking. Could handle subsequently if you prefer.

recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,404 @@
...
+            // For cases like: const { builtinModules } = await import('node:repl');
+            const originalText = objectPattern.text();
+            if (containsBuiltinProperties(originalText)) {
+                if (originalText.includes("_builtinLibs")) {
+                    const newText = originalText.replace("_builtinLibs", "builtinModules");
+                    edits.push(objectPattern.replace(newText));
+                    replaceStandaloneBuiltinLibsReferences(rootNode, edits);
+                }
+            }
+        } else {
```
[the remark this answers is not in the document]
AugustinMauroy: idk what does it's mean

recipes/repl-builtin-modules/src/workflow.ts
```diff
@@ -0,0 +1,404 @@
...
+                    rule: { kind: "shorthand_property_identifier_pattern" }
+                });
+                const pairProperties = objectPattern.findAll({
+                    rule: { kind: "pair_pattern" }
+                });
+
+				// Check if only builtin properties are destructured
+				const builtinShorthandCount = properties.filter(p =>
+					containsBuiltinProperties(p.text())
+				).length;
```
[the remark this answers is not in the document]
JakobJingleheimer: There's not enough indentation or maybe is a tabs vs spaces thing for these lines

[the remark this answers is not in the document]
AugustinMauroy: oh get it

PR #151, opened 2025-07-30
feat(util): add utility function to resolve binding path

brunocroh:
## Description

So, while looking through the recipes developed and the new ones that will need to be developed, a very common problem among many of them is resolving the binding path per file.

i.e.: if the codemod needs to replace

```js
import { types } from 'node:util';

if (types.isNativeError(err)) {
  // handle the error
}
```
with

```js
if (Error.isError(err)) {
  // handle the error
}
```

A very tricky case to handle is how it was imported in that file, because there are many ways to do it. The ones I can remember now are:

```js
import utilImport from "node:util";
import { types as renameIt } from "node:util";
import * as util from "node:util";
import anyNameHere from "util";

const utils = require("node:util");
var { types: typesRequire } = require("node:util");
let { types: { isNativeError } } = require("util");

/// usage
isNativeError(err)
typesRequire.isNativeError(err)
utilImport.types.isNativeError(err)
util.types.isNativeError(err)
anyNameHere.types.isNativeError(err)
utils.types.isNativeError(err)
renameIt.types.isNativeError(err)
```

So the idea is that the utility function should be able to check all these scenarios and return how the function will be used in this context, for example:

```js
// import utilImport from "node:util";
resolveBindingPath(nodeWithImport, "$.types.isNativeError")
// result: utilImport.types.IsNativeError

// const { types: { isNativeError } } = require('node:util');
resolveBindingPath(nodeWithImport, "$.types.isNativeError")
// result: isNativeError
```

What do you guys think about this API?
Did I miss any import/require scenario? (Check the test file for all examples this function is currently supporting.)

### DIFF
--- a/utils/src/ast-grep/resolve-binding-path.test.ts
+++ b/utils/src/ast-grep/resolve-binding-path.test.ts
@@ -54,8 +54,6 @@ describe("resolve-binding-path", () => {
 
 		const bindingPath = resolveBindingPath(requireStatement!, "$.types.isNativeError");
 
-		console.log({ bindingPath });
-
 		assert.strictEqual(bindingPath, "isNativeError");
 	});
 
@@ -227,9 +225,9 @@ describe("resolve-binding-path", () => {
 		assert.strictEqual(bindingPath, "renamed.isNativeError");
 	});
 
-	it("should handle empty path segments gracefully", () => {
+	it("should handle deep destructuring and return remaining path after resolved binding", () => {
 		const code = dedent`
-			const { types } = require('node:util');
+			const { a: {b: { c: { d }} } } = require('node:util');
 		`;
 
 		const rootNode = astGrep.parse(astGrep.Lang.JavaScript, code);
@@ -239,9 +237,9 @@ describe("resolve-binding-path", () => {
 			},
 		});
 
-		const bindingPath = resolveBindingPath(requireStatement!, "$.types");
+		const bindingPath = resolveBindingPath(requireStatement!, "$.a.b.c.d.e");
 
-		assert.strictEqual(bindingPath, "types");
+		assert.strictEqual(bindingPath, "d.e");
 	});
 
 	it("should handle single character variable names", () => {


utils/src/ast-grep/resolve-binding-path.ts
```diff
@@ -0,0 +1,226 @@
+import type { TypesMap } from "@codemod.com/jssg-types/main";
+import type { Kinds } from "@ast-grep/napi/types/staticTypes.js";
```
AugustinMauroy, line 2: what if we only use `@codemod.com/jssg-types` ?

utils/src/ast-grep/resolve-binding-path.test.ts
```diff
@@ -0,0 +1,151 @@
+import assert from "node:assert/strict";
+import { resolveBindingPath } from "./resolve-binding-path.ts";
+import { describe, it, only } from "node:test";
+import astGrep from "@ast-grep/napi";
+import dedent from "dedent";
```
AugustinMauroy, line 5:
```suggestion
import assert from "node:assert/strict";
import { describe, it, only } from "node:test";
import astGrep from "@ast-grep/napi";
import dedent from "dedent";
import { resolveBindingPath } from "./resolve-binding-path.ts";
```

my import style guide

- node 
- npm apckage
- relative path 
- type import with same order

AugustinMauroy (commented):
good starting point

utils/src/ast-grep/resolve-binding-path.ts
brunocroh: my bad, replaced by `@ast-grep/napi` imports

utils/src/ast-grep/resolve-binding-path.test.ts
```diff
@@ -0,0 +1,159 @@
...
+		`;
+
+		const rootNode = astGrep.parse(astGrep.Lang.JavaScript, code);
+		const functionDeclaration = rootNode.root().find({
+			rule: {
+				kind: "function_declaration",
+			},
+		});
+
+		assert.throws(() => resolveBindingPath(functionDeclaration!, "$.types.isNativeError"));
```
AugustinMauroy, line 89: why not just putting null and handling at codemod level ?

The question is deeper that just here. Should our utility throw error or should return null and handled in codemod ?

My tought is throw error but document it so we can catch it as codemod level

AugustinMauroy (commented):
LGTM ! Check if we can use in on `recipes/**` to simplify our current codemod

utils/src/ast-grep/resolve-binding-path.test.ts
brunocroh: My idea behind throwing an error was to make development easier, to alert whoever is working on it that the node being passed is incorrect in some way. But yeah, I agree with you, it does make the usage a bit awkward.

I'll go ahead and remove those throw scenarios.

AugustinMauroy: Let's keep it !

AugustinMauroy:
None of our recipes/codemods support `= await import(‘node:’)`. Should we support it in this utility?

brunocroh:
> None of our recipes/codemods support `= await import(‘node:’)`. Should we support it in this utility?
> 
> IMO if the import have the await statement yes instead we will not try to understand pending promises

The idea of this utility function is not to be identifying the import/require statement, but is resolve the binding (name of functions, vars on this file context).

So the idea is got the require/import (maybe just a simple destructuring) and pass to it return the path of the function/var that will be replaced in the file.

So it can be used combined with utility functions to return the imported nodes like

```js
	const requireDeclarations = getNodeRequireCalls(root, "process"); // utility that already exist in the project
	for (const declarations of requireDeclarations) {
	  const referencesInTheFile = resolveBindingPath(declarations, "$.mainModule")
	  const itemsToReplace = root.findAll({
	    rule: {
	      regex: referencesInTheFile
	    }
	  })
	}
	
	for (const item of itemsToReplace){
	  //implement replace
	}

```

AugustinMauroy:
We need a second approval from @nodejs/userland-migrations

brunocroh:
> Hello 👋, just a quick heads up is this utility function planned to be merged ? I was working on a issue and realized this could be quite useful for my case as well. Since the PR hasn't been merged yet, I just wanted to check if it's still being considered. Thanks

Yep, the same thing happened to me. What I did was copy and paste the code into another branch and develop using it. Just make sure not to include it in your commits. Once it's merged, you can rebase on your side.

JakobJingleheimer (commented):
Thanks! This looks super useful 🙌 Just a few minor items.

utils/src/ast-grep/resolve-binding-path.ts
```diff
@@ -0,0 +1,256 @@
...
+		const text = namedImport.text();
+		if (pathArr.includes(text)) {
+			importFound = text;
+			break;
+		}
+	}
+
+	if (importFound) {
+		return path.slice(path.indexOf(importFound));
+	}
```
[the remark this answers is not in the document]
brunocroh: Good catch, better now.

JakobJingleheimer (approved):
🙌 w00t—thanks!

utils/src/ast-grep/resolve-binding-path.test.ts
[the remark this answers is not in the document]
JakobJingleheimer: Thanks!

AugustinMauroy:
@0xlakshan go ahead !

PR #154, opened 2025-08-03
feat(util): add removeBinding and refactor recipes to use it

brunocroh:
Adds `removeBinding`, a new utility function to remove bindings when they are no longer used in the file, and refactors `recipe/process-main-module` to use this function.

**The responsibility of checking whether a binding is used or not does not belong to this function.** This function simply returns an object that indicates what needs to be done to the require/import statements.

The function return type is shown below:

```typescript
import type { Edit, Range } from "@ast-grep/napi";

type RemoveBindingReturnType = {
	edit?: Edit;
	lineToRemove?: Range;
};
```

So an example of usage will be

```typescript
const result = removeBinding(declarationNode, "mainModule");

if (result?.edit) {
	edits.push(result.edit);
}

if (result?.lineToRemove) {
	linesToRemove.push(result.lineToRemove);
}
```

### CommonJS Require Statements

| Scenario | Code Example | Call | Expected Result |
|----------|--------------|------|-----------------|
| **Single require binding** | `const util = require('node:util');` | `removeBinding(node, 'util')` | **Remove entire line** |
| **Single require binding (no match)** | `const util = require('node:util');` | `removeBinding(node, 'types')` | **Return undefined** |
| **Single destructured require** | `const { types } = require('node:util');` | `removeBinding(node, 'types')` | **Remove entire line** |
| **Multiple destructured require** | `const { types, diff } = require('node:util');` | `removeBinding(node, 'types')` | **Edit:** → `const { diff } = require('node:util');` |
| **Single destructured from object** | `const { mainModule } = process;` | `removeBinding(node, 'mainModule')` | **Remove entire line** |

### ES6 Import Statements

| Scenario | Code Example | Call | Expected Result |
|----------|--------------|------|-----------------|
| **Default import** | `import util from 'node:util';` | `removeBinding(node, 'util')` | **Remove entire line** |
| **Default import (no match)** | `import util from 'node:util';` | `removeBinding(node, 'types')` | **Return undefined** |
| **Namespace import** | `import * as util from 'node:util';` | `removeBinding(node, 'util')` | **Remove entire line** |
| **Namespace import (no match)** | `import * as util from 'node:util';` | `removeBinding(node, 'types')` | **Return undefined** |
| **Single named import** | `import { types } from 'node:util';` | `removeBinding(node, 'types')` | **Remove entire line** |
| **Multiple named imports** | `import { types, diff } from 'node:util';` | `removeBinding(node, 'types')` | **Edit** → `import { diff } from 'node:util';` |
| **Named import (no match)** | `import { types, diff } from 'node:util';` | `removeBinding(node, 'none')` | **Return undefined** |
| **Single aliased import** | `import { types as utilTypes } from 'node:util';` | `removeBinding(node, 'utilTypes')` | **Remove entire line** |
| **Mixed aliased imports** | `import { types as utilTypes, diff } from 'node:util';` | `removeBinding(node, 'utilTypes')` | **Edit:** → `import { diff } from 'node:util';` |
| **Multiple aliased imports** | `import { types as utilTypes, diff as utilDiffs } from 'node:util';` | `removeBinding(node, 'utilTypes')` | **Edit:** → `import { diff as utilDiffs } from 'node:util';` |

### DIFF
--- a/recipes/process-main-module/codemod.yaml
+++ b/recipes/process-main-module/codemod.yaml
@@ -1,6 +1,6 @@
 schema_version: "1.0"
 name: "@nodejs/process-main-module"
-version: 1.1.0
+version: 1.0.1
 description: Handle DEP0138 via transforming `process.mainModule` to `require.main`.
 author: Bruno Rodrigues
 license: MIT


recipes/correct-ts-specifiers/src/fixtures/e2e/test.ts
AugustinMauroy, line 1: this shouldn't commit (this codemod was written with legacy codemod cli and it's doesn't have test thing it's why we need to restore it)

AugustinMauroy (approved):
LGTM ! super clean

recipes/process-main-module/src/workflow.ts
AugustinMauroy, line 1: bump minor of this codemod se we can release it after this pr

AugustinMauroy:
BTW red ci si handle in #155

recipes/correct-ts-specifiers/src/fixtures/e2e/test.ts
brunocroh: I didn't notice that change, my bad

recipes/process-main-module/src/workflow.ts
brunocroh: done

brunocroh:
> also `rmdir` can profit of it

I don't think so, because the `rmdir` recipe replaces the old functions with new ones, and `removeBindings` is for when you need to remove something that will no longer be useful.

Issue #123, opened 2025-07-25
feat: handle `crypto.createCredentials()` depreciation


AugustinMauroy:
## Description

Since `crypto.createCredentials()` is deprecated (DEP0010) and has reached End-of-Life status, we should provide a codemod to replace it.

- The codemod should replace all instances of `crypto.createCredentials()` with `tls.createSecureContext()`.
- The codemod should update import/require statements to replace `node:crypto` with `node:tls` when `createCredentials` is the only crypto function being used.
- If the `node:crypto` import is destructured and only contains `createCredentials`, it should be replaced with `node:tls` destructured import.
- If there are other crypto functions being used, the codemod should add a separate `node:tls` import and keep the existing `node:crypto` import.

## Additional Information

Note that `crypto.createCredentials()` was removed in Node.js v11.0.0 and replaced with `tls.createSecureContext()`. The functionality is identical, but the API has been moved to the more appropriate `tls` module. The options object format remains the same (key, cert, ca, etc.).

### Case 1: Destructured import with only createCredentials

**Before:**
```js
const { createCredentials } = require('node:crypto');

const credentials = createCredentials({
  key: privateKey,
  cert: certificate,
  ca: [caCertificate]
});
```

**After:**
```js
const { createSecureContext } = require('node:tls');

const credentials = createSecureContext({
  key: privateKey,
  cert: certificate,
  ca: [caCertificate]
});
```

### Case 2: Namespace import

**Before:**
```js
const crypto = require('node:crypto');

const credentials = crypto.createCredentials({
  key: fs.readFileSync('server-key.pem'),
  cert: fs.readFileSync('server-cert.pem')
});
```

**After:**
```js
const tls = require('node:tls');

const credentials = tls.createSecureContext({
  key: fs.readFileSync('server-key.pem'),
  cert: fs.readFileSync('server-cert.pem')
});
```

### Case 3: Mixed usage with other crypto functions

**Before:**
```js
const { createCredentials, createHash } = require('node:crypto');

const credentials = createCredentials({
  key: privateKey,
  cert: certificate
});

const hash = createHash('sha256');
hash.update('some data');
```

**After:**
```js
const { createHash } = require('node:crypto');
const { createSecureContext } = require('node:tls');

const credentials = createSecureContext({
  key: privateKey,
  cert: certificate
});

const hash = createHash('sha256');
hash.update('some data');
```

### Case 4: ESM import

**Before:**
```js
import { createCredentials } from 'node:crypto';

[...]

PR #156, opened 2025-08-03
feat(tls.createSecureContext): introduce

0hmX:
closes: #123

### DIFF
--- a/recipes/crypto-createCredentials/README.md
+++ b/recipes/crypto-createCredentials/README.md
@@ -1,0 +1,51 @@
+# `fs.rmdir` DEP0147
+
+This recipe provides a guide for migrating from the deprecated `crypto.createCredentials` method in Node.js.
+
+See [DEP0010](https://nodejs.org/api/deprecations.html#DEP0010).
+
+## Examples
+
+**Before:**
+
+```js
+// Using the deprecated createCredentials from node:crypto
+const { createCredentials } = require('node:crypto');
+
+const credentials = createCredentials({
+  key: privateKey,
+  cert: certificate,
+  ca: [caCertificate]
+});
+
+// Using destructuring with ES module imports
+import { createCredentials } from 'node:crypto';
+
+const credentials = createCredentials({
+  key: privateKey,
+  cert: certificate,
+  ca: [caCertificate]
+});
+```
+
+**After:**
+
+```js
+// Updated to use createSecureContext from node:tls
+const { createSecureContext } = require('node:tls');
+
+const credentials = createSecureContext({
+  key: privateKey,
+  cert: certificate,
+  ca: [caCertificate]
+});
+
+// Updated ES module import
+import { createSecureContext } from 'node:tls';
+
+const credentials = createSecureContext({
+  key: privateKey,
+  cert: certificate,
+  ca: [caCertificate]
+});
+```
--- a/recipes/crypto-createCredentials/codemod.yaml
+++ b/recipes/crypto-createCredentials/codemod.yaml
@@ -1,0 +1,21 @@
+schema_version: "1.0"
+name: "@nodejs/crypto-createCredentials"
+version: 1.0.0
+description: Handle DEP0010 via transforming `crypto.createCredentials` to `tls.createSecureContext`
+author: 0hmx
+license: MIT
+workflow: workflow.yaml
+category: migration
+
+targets:
+  languages:
+    - javascript
+    - typescript
+
+keywords:
+  - transformation
+  - migration
+
+registry:
+  access: public
+  visibility: public
--- a/recipes/crypto-createCredentials/package.json
+++ b/recipes/crypto-createCredentials/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "@nodejs/crypto-createCredentials",
+  "version": "1.0.0",
+  "description": "Handle DEP0010 via transforming `crypto.createCredentials` to `tls.createSecureContext` with the appropriate options.",
+  "type": "module",
+  "scripts": {
+    "test": "npx codemod@next jssg test -l typescript ./src/workflow.ts ./"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/nodejs/userland-migrations.git",
+    "directory": "recipes/rmdirs",
+    "bugs": "https://github.com/nodejs/userland-migrations/issues"
+  },
+  "author": "0hmx",
+  "license": "MIT",
+  "homepage": "https://github.com/nodejs/userland-migrations/blob/main/recipes/rmdirs/README.md",
+  "devDependencies": {
+    "@types/node": "^24.0.3",
+    "@codemod.com/jssg-types": "^1.0.3"
+  },
+  "dependencies": {
+    "@nodejs/codemod-utils": "*"
+  }
+}
--- a/recipes/crypto-createCredentials/workflow.yaml
+++ b/recipes/crypto-createCredentials/workflow.yaml
@@ -1,0 +1,25 @@
+version: "1"
+
+nodes:
+  - id: apply-transforms
+    name: Apply AST Transformations
+    type: automatic
+    runtime:
+      type: direct
+    steps:
+      - name: Replace `assert` import attribute to the `with` ECMAScript import attribute.
+        js-ast-grep:
+          js_file: src/workflow.ts
+          base_path: .
+          include:
+            - "**/*.js"
+            - "**/*.jsx"
+            - "**/*.mjs"
+            - "**/*.cjs"
+            - "**/*.cts"
+            - "**/*.mts"
+            - "**/*.ts"
+            - "**/*.tsx"
+          exclude:
+            - "**/node_modules/**"
+          language: typescript
--- a/recipes/crypto-createCredentials/src/workflow.ts
+++ b/recipes/crypto-createCredentials/src/workflow.ts
@@ -1,0 +1,328 @@
+import type { SgRoot, Edit, SgNode } from "@codemod.com/jssg-types/main";
+
+type ImportType =
+    | 'DESTRUCTURED_REQUIRE'
+    | 'NAMESPACE_REQUIRE'
+    | 'DESTRUCTURED_IMPORT'
+    | 'NAMESPACE_IMPORT';
+
+function ensureTlsImport(
+    rootNode: SgNode,
+    edits: Edit[],
+    importType: ImportType,
+    tlsImportState: { added: boolean; identifier: string },
+): string {
+    if (tlsImportState.added) {
+        return tlsImportState.identifier;
+    }
+
+    const isEsm =
+        importType === 'DESTRUCTURED_IMPORT' || importType === 'NAMESPACE_IMPORT';
+    const moduleSpecifier = 'node:tls';
+
+    const findRule = {
+        rule: {
+            kind: isEsm ? 'import_statement' : 'variable_declarator',
+            has: {
+                field: isEsm ? 'source' : 'value',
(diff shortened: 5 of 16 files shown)


recipes/crypto-createCredentials/README.md
```diff
@@ -0,0 +1,51 @@
+# `fs.rmdir` DEP0147
```
AugustinMauroy, line 1: Copy pasta

recipes/crypto-createCredentials/workflow.yaml
```diff
@@ -0,0 +1,25 @@
+version: "1"
+
+nodes:
+  - id: apply-transforms
+    name: Apply AST Transformations
+    type: automatic
+    runtime:
+      type: direct
+    steps:
+      - name: Replace `assert` import attribute to the `with` ECMAScript import attribute.
```
AugustinMauroy, line 10: Copy pasta

AugustinMauroy (commented):
Minor thing

recipes/crypto-createCredentials/codemod.yaml
```diff
@@ -0,0 +1,21 @@
+schema_version: "1.0"
+name: "@nodejs/crypto-createCredentials"
```
brunocroh, line 2:
```suggestion
name: "@nodejs/crypto-create-credentials"
```

@AugustinMauroy I think we are using only kebab-case for codemod recipes names, right? 

I think we're using only kebab-case for codemod recipe names, right?

It would be great to have a section in `CONTRIBUTE.md` about how to name codemod recipes and the naming convention as well.

recipes/crypto-createCredentials/package.json
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/nodejs/userland-migrations.git",
+    "directory": "recipes/rmdirs",
+    "bugs": "https://github.com/nodejs/userland-migrations/issues"
+  },
+  "author": "0hmx",
+  "license": "MIT",
+  "homepage": "https://github.com/nodejs/userland-migrations/blob/main/recipes/rmdirs/README.md",
```
brunocroh, line 17:
```suggestion
  "homepage": "https://github.com/nodejs/userland-migrations/blob/main/recipes/crypto-create-credentials/README.md",
```

recipes/crypto-createCredentials/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "devDependencies": {
+    "@types/node": "^24.0.3",
```
brunocroh, line 19:
```suggestion
```

Actually no needed since the codemod run on jssg context not it node.js

recipes/crypto-createCredentials/codemod.yaml
brunocroh: @0hmX just to you know this name `crypto-create-credentials` matches the recipe folder name, so you could you please update the folder name as well? cheers

0hmX: Sure, I was also thinking the same

recipes/crypto-createCredentials/README.md
0hmX: my bad

recipes/crypto-create-credentials/src/workflow.ts
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    tlsImportState.added = true;
+    return tlsImportState.identifier;
+}
+
+function handleDestructuredImport(
+    statement: SgNode,
+    rootNode: SgNode,
+    edits: Edit[],
```
brunocroh, line 64: What do you think about having all utility functions return `Edit[]` arrays, and then applying these edits only within the `transform` function itself by adding them to a list of rules and committing the changes to the file?

For example:
```js
const edits = handleDestructuredImport(/* pass needed props */)
allEdits = allEdits.concat(edits)
rootNode.commitEdits(allEdits)
```

I think this approach would make the code easier to follow, as all mutable changes would happen in the top-level context (the transform function body), while all other functions would just read data and return it (making them pure functions).

brunocroh: In case you need to return more data, you can return something like:

```typescript
type ExampleReturnType = {
  edits: Edit[]
  someBoolValue: boolean
}
```

recipes/crypto-create-credentials/workflow.yaml
```diff
@@ -0,0 +1,25 @@
+version: "1"
```
AugustinMauroy, line 1:
```suggestion
# yaml-language-server: $schema=https://raw.githubusercontent.com/codemod-com/codemod/refs/heads/main/schemas/workflow.json

version: "1"
```

recipes/crypto-create-credentials/workflow.yaml
```diff
@@ -0,0 +1,25 @@
...
+          base_path: .
+          include:
+            - "**/*.js"
+            - "**/*.jsx"
+            - "**/*.mjs"
+            - "**/*.cjs"
+            - "**/*.cts"
+            - "**/*.mts"
+            - "**/*.ts"
+            - "**/*.tsx"
```
brunocroh, line 22: nit: Alphabetical order
```suggestion
            - "**/*.cjs"
            - "**/*.js"
            - "**/*.jsx"
            - "**/*.mjs"
            - "**/*.cts"
            - "**/*.mts"
            - "**/*.ts"
            - "**/*.tsx"
```

recipes/crypto-create-credentials/src/workflow.ts
0hmX: Agreed, this pattern does look better

recipes/crypto-create-credentials/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+This recipe transforms `crypto.createCredentials` usage to use modern `node:tls` methods.
+
+See [DEP0010](https://nodejs.org/api/deprecations.html#DEP0010).
+
+## Examples
+
+**Before:**
+
+```js
```
AugustinMauroy, line 11:
```suggestion
**Before:**
```js
```

recipes/crypto-create-credentials/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "description": "Handle DEP0010 via transforming `crypto.createCredentials` to `tls.createSecureContext` with the appropriate options.",
+  "type": "module",
+  "scripts": {
+    "test": "npx codemod@next jssg test -l typescript ./src/workflow.ts ./"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/nodejs/userland-migrations.git",
+        "directory": "recipes/crypto-create-credentials",
+    "bugs": "https://github.com/nodejs/userland-migrations/issues"
```
AugustinMauroy, line 13:
```suggestion
    "url": "git+https://github.com/nodejs/userland-migrations.git",
    "directory": "recipes/crypto-create-credentials",
    "bugs": "https://github.com/nodejs/userland-migrations/issues"
```

AugustinMauroy (approved):
LGMT ! Thanks for you first contribution.
If you have any feedback about DX on this project let's me know. And if you have any feedback about codemod CLI/platform I'm going to transfer it to them

AugustinMauroy:
we also have few utility you should use it to simplify

recipes/crypto-create-credentials/src/workflow.ts
```diff
@@ -0,0 +1,397 @@
...
+            .filter((id) => id.text() !== 'createCredentials')
+            .map((id) => id.text());
+
+        const newDestructuredString = `{ ${otherProps.join(', ')} }`;
+        localEdits.push(specifiersNode.replace(newDestructuredString));
+
+        if (!currentState.added) {
+            const newEdit = {
+                startPos: statement.range().end.index,
+                endPos: statement.range().end.index,
```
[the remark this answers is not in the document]
0hmX: Yes, we are adding it at the end. So, start and end both are pointing to the end of the statement.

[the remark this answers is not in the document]
JakobJingleheimer: I think a brief code comment about that would be a kindness to our future selves :)

```suggestion
                startPos: statement.range().end.index,
                // appending, so use end for both
                endPos: statement.range().end.index,
```

recipes/crypto-create-credentials/workflow.yaml
JakobJingleheimer: @AugustinMauroy we should probably add a CI check (in another PR) that `yaml-language-server` is present for certain files (and then if it isn't correct, we have another check that will fail—but that one won't fail when this is missing missing).

JakobJingleheimer (commented):
Looks like good test cases :)

Regarding cases handled, what about import expressions `const { createCredentials } = await import('node:crypto')` and variants like `const createCredentials = (await import('node:crypto')).then((m) => m.createCredentials)`?

I think that is needed in order to say the deprecation is handled.

AugustinMauroy:
> > BTW any of our publish codemod support dynamic import
> 
> Sorry, is that a question or a statement? (Do you mean "all of our published codemods support dynamic import"?)

It's a fact ! So do we want to support that in general ?

recipes/crypto-create-credentials/src/workflow.ts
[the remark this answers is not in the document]
AugustinMauroy: > Not necessarily a blocker, but this file has a lot more code than I expected. There are basically 3 things that need handling, 2 of which are mostly the same:

Also it's because it's doesn't use any utility.

@0hmX what do you think about using our utility to reduce the amount of code here ?

0hmX:
Okay, so I ended up rewriting the code, but it's now bigger than the previous version🫠. Yes, I am using the utilities. I added a few more test cases related to pair_pattern and import_specifier.

However, I couldn't get the logic to work for `const createCredentials = (await import('node:crypto')).then((m) => m.createCredentials)` because the getNodeImportCalls utility does not handle this case. I want to skip it, as adding the extra code would make the file too large.

recipes/createCredentials-to-createSecureContext/src/workflow.ts
```diff
@@ -0,0 +1,370 @@
...
+                    { has: { field: 'object', regex: `^${localNamespace}$` } },
+                    { has: { field: 'property', regex: `^${oldFunctionName}$` } }
+                ]
+            }
+        }
+    });
+
+    if (usages.length === 0) {
+        return [];
+    }
```
AugustinMauroy, line 34:
```suggestion
    if (usages.length === 0) return [];
```

recipes/createCredentials-to-createSecureContext/src/workflow.ts
```diff
@@ -0,0 +1,370 @@
...
+    for (const node of nodes || []) {
+      const edits = handler(node, root);
+      if (edits.length) {
+        wasTransformed = true;
+        allEdits.push(...edits);
+      }
+    }
+  }
+
+  return wasTransformed ? rootNode.commitEdits(allEdits) : null;
```
AugustinMauroy, line 369: wasTranformed var and be replaced by `!allEdits.lenght`.

Also use this structure instead of ternary:

```js
if (!allEdits.lenght) return null

return rootNode.commitEdits(allEdits)
```

AugustinMauroy:
I'm not sure to know what happened to your `package-lock.json` try `npm I` first then if it's doesn't work `npm update`

AugustinMauroy:
I'm going to take over this pr so we can land it

brunocroh:
LGTM

PR #157, opened 2025-08-03
feat(types-is-native-error): introduce

### LINKED ISSUE #113: feat: handle `util.types.isNativeError()`  depreciation
## Description

Since `util.types.isNativeError()` is deprecated in Node.js v24.2.0, we should provide a codemod to replace it.

- The codemod should remove import if it's only thing imported or remove the member if not anymore used.
- The codemod should modify any call expression of this api.

## Examples

**Before:**

```js
import { types } from 'node:util';


if (types.isNativeError(err)) {
  // handle the error
}


```

**After:**

```js
if (Error.isError(err)) {
  // handle the error
}
```

## REFS

- [DEP0197: `util.types.isNativeError()`](https://nodejs.org/docs/latest/api/deprecations.html#dep0197-utiltypesisnativeerror)

brunocroh:
Closes #113

This PR depends on PR #151 being merged first.

## Overview
This PR introduces a new TypeScript transformation that automatically migrates deprecated `util.types.isNativeError` usage to the modern `Error.isError` API.

## Additional work

This implements one additional case for `utils/src/ast-grep/remove-binding.ts`. Previously, it failed to remove deep named imports, so I fixed this behaviour and added a new test to prevent regression

### Previous behaviour:
input 
```js
const { types: { isNativeError, isMap } } = require("util");
 ```
  
result
 ```js
 const { isMap } = require("util");
 ```

### New behaviour:
input 
```js
const { types: { isNativeError, isMap } } = require("util");
 ```
  
result
 ```js
const { types: { isMap } } = require("util");
 ```

### DIFF
--- a/recipes/types-is-native-error/src/workflow.ts
+++ b/recipes/types-is-native-error/src/workflow.ts
@@ -141,6 +141,6 @@ export default function transform(root: SgRoot<JS>): string | null {
 		}
 	}
 
-	let sourceCode = rootNode.commitEdits(edits);
+	const sourceCode = rootNode.commitEdits(edits);
 	return removeLines(sourceCode, linesToRemove);
 }
--- a/recipes/types-is-native-error/src/workflow.ts
+++ b/recipes/types-is-native-error/src/workflow.ts
@@ -1,5 +1,4 @@
 import type { Edit, Range, SgNode, SgRoot } from "@codemod.com/jssg-types/main";
-import type JS from "@codemod.com/jssg-types/langs/javascript";
 import { getNodeRequireCalls } from "@nodejs/codemod-utils/ast-grep/require-call";
 import { getNodeImportStatements } from "@nodejs/codemod-utils/ast-grep/import-statement";
 import { resolveBindingPath } from "@nodejs/codemod-utils/ast-grep/resolve-binding-path";
@@ -48,15 +47,6 @@ function createPropBinding(
 	const lastPropertyAccess = pathArr.at(-1);
 	const propertyAccess = pathArr.slice(0, -1).join(".");
 
-	if (!propertyAccess) {
-		return {
-			path,
-			propertyAccess,
-			lastPropertyAccess,
-			depth: pathArr.length,
-		};
-	}
-
 	return {
 		path,
 		propertyAccess,
@@ -82,15 +72,13 @@ function createPropBinding(
  *    isNativeError have been replaced
  *
  */
-export default function transform(root: SgRoot<JS>): string | null {
+export default function transform(root: SgRoot): string | null {
 	const rootNode = root.root();
 	const bindings: Binding[] = [];
 	const edits: Edit[] = [];
 	const linesToRemove: Range[] = [];
 
-	// @ts-expect-error - ast-grep types are not fully compatible with JSSG types
 	const nodeRequires = getNodeRequireCalls(root, "util");
-	// @ts-expect-error - ast-grep types are not fully compatible with JSSG types
 	const nodeImports = getNodeImportStatements(root, "util");
 	const path = "$.types.isNativeError";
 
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: recipes/types-is-native-error/workflow.yaml [hunk 1]
```diff
@@ -0,0 +1,26 @@
+version: "1"
```
AugustinMauroy, line 1:
```suggestion
# yaml-language-server: $schema=https://raw.githubusercontent.com/codemod-com/codemod/refs/heads/main/schemas/workflow.json

version: "1"
```

AugustinMauroy (approved):
CI is green.

recipes/types-is-native-error/src/workflow.ts
```diff
@@ -0,0 +1,146 @@
...
+ * @example
+ * ```typescript
+ * removeLastPropertyAccess("foo.bar.baz");
+ * // Returns: { path: "foo.bar.baz", propertyAccess: "foo.bar", lastPropertyAccess: "baz", depth: 3 }
+ *
+ * removeLastPropertyAccess("foo");
+ * // Returns: { path: "foo", propertyAccess: "", lastPropertyAccess: "foo", depth: 1 }
+ * ```
+ */
+function removeLastPropertyAccess(
```
avivkeller, line 36: nit: I think this title doesn't accurately describe the function.

Can we make this a createPropBinding function that calls a removeLastPropertyAccess which specifically does the string transformation?

recipes/types-is-native-error/src/workflow.ts
```diff
@@ -0,0 +1,146 @@
...
+
+	if (!pathArr) {
+		return {
+			path,
+			depth: 1,
+		};
+	}
+
+	const lastPropertyAccess = pathArr.at(-1);
+	const propertyAccess = pathArr.slice(0, pathArr.length - 1).join(".");
```
JakobJingleheimer, line 49: I think you can just pass `-1` to the second arg and it will grab until but not including the last character.

```suggestion
	const propertyAccess = pathArr.slice(0, -1).join(".");
```

recipes/types-is-native-error/src/workflow.ts
```diff
@@ -0,0 +1,146 @@
...
+			depth: pathArr.length,
+		};
+	}
+
+	return {
+		path,
+		propertyAccess,
+		lastPropertyAccess,
+		depth: pathArr.length,
+	};
```
JakobJingleheimer, line 65: These are the same O.o

recipes/types-is-native-error/src/workflow.ts
```diff
@@ -0,0 +1,146 @@
...
+ */
+export default function transform(root: SgRoot<JS>): string | null {
+	const rootNode = root.root();
+	const bindings: Binding[] = [];
+	const edits: Edit[] = [];
+	const linesToRemove: Range[] = [];
+
+	// @ts-ignore - ast-grep types are not fully compatible with JSSG types
+	const nodeRequires = getNodeRequireCalls(root, "util");
+	// @ts-ignore - ast-grep types are not fully compatible with JSSG types
```
JakobJingleheimer, line 93: So we get alerted when it's fixed and this can be adjusted/removed.

```suggestion
	// @ts-expect-error - ast-grep types are not fully compatible with JSSG types
	const nodeRequires = getNodeRequireCalls(root, "util");
	// @ts-expect-error - ast-grep types are not fully compatible with JSSG types
```

recipes/types-is-native-error/src/workflow.ts
```diff
@@ -0,0 +1,146 @@
...
+			},
+		});
+
+		for (const node of nodesToEdit) {
+			edits.push(node.replace("Error.isError"));
+		}
+
+		if (nodes.length === nodesToEdit.length) {
+			const result = removeBinding(
+				// @ts-ignore - ast-grep types are not fully compatible with JSSG types
```
JakobJingleheimer, line 129:
```suggestion
				// @ts-expect-error - ast-grep types are not fully compatible with JSSG types
```

JakobJingleheimer (commented):
Looks pretty good. I think the if … return A else return A might be a mistake?

Annnd broken record: dynamic imports

brunocroh:
> Looks pretty good. I think the if … return A else return A might be a mistake?
> 
> Annnd broken record: dynamic imports

All done

recipes/types-is-native-error/src/workflow.ts
brunocroh: Isn't necessary any more, so removed

JakobJingleheimer (approved):
🙌

PR #164, opened 2025-08-05
feat(crypto): add RSA-PSS codemod for DEP0154

### LINKED ISSUE #144: Feat: handle RSA-PSS generate key pair options
## Description

Since RSA-PSS generate key pair options are deprecated (DEP0154) and have reached End-of-Life status, we should provide a codemod to replace them.

- The codemod should replace deprecated `'hash'` option with `'hashAlgorithm'` in RSA-PSS key generation functions.
- The codemod should replace deprecated `'mgf1Hash'` option with `'mgf1HashAlgorithm'` in RSA-PSS key generation functions.
- The codemod should handle both `crypto.generateKeyPair()` and `crypto.generateKeyPairSync()` functions.
- The codemod should only transform options objects for RSA-PSS key type (`'rsa-pss'`).

## Additional Information

Note that the deprecated options `'hash'` and `'mgf1Hash'` were removed in a recent Node.js version. The options were deprecated to provide more descriptive and consistent naming. The `'hash'` option should be replaced with `'hashAlgorithm'` and `'mgf1Hash'` should be replaced with `'mgf1HashAlgorithm'`.

This deprecation only affects RSA-PSS key pair generation and does not impact other key types or crypto operations.

### Case 1: crypto.generateKeyPair with hash option

**Before:**
```js
const crypto = require('node:crypto');

crypto.generateKeyPair('rsa-pss', {
  modulusLength: 2048,
  hash: 'sha256',
  saltLength: 32
}, (err, publicKey, privateKey) => {
  // handle keys
});
```

**After:**
```js
const crypto = require('node:crypto');

crypto.generateKeyPair('rsa-pss', {
  modulusLength: 2048,
  hashAlgorithm: 'sha256',
  saltLength: 32
}, (err, publicKey, privateKey) => {
  // handle keys
});
```

### Case 2: crypto.generateKeyPairSync with mgf1Hash option

**Before:**
```js
const crypto = require('node:crypto');

const { publicKey, privateKey } = crypto.generateKeyPairSync('rsa-pss', {
  modulusLength: 2048,
  mgf1Hash: 'sha256',
  saltLength: 32
});
```

**After:**
```js
const crypto = require('node:crypto');

const { publicKey, privateKey } = crypto.generateKeyPairSync('rsa-pss', {
  modulusLength: 2048,
  mgf1HashAlgorithm: 'sha256',
  saltLength: 32
});
```

### Case 3: Both deprecated options together

**Before:**
```js
const crypto = require('node:crypto');

crypto.generateKeyPair('rsa-pss', {
  modulusLength: 4096,
  hash: 'sha512',
  mgf1Hash: 'sha512',
  saltLength: 64,
  publicKeyEncoding: {
    type: 'spki',
    format: 'pem'
  },
  privateKeyEncoding: {
    type: 'pkcs8',
    format: 'pem'
  }
}, callback);
```

**After:**
```js
const crypto = require('node:crypto');

crypto.generateKeyPair('rsa-pss', {
  modulusLength: 4096,
  hashAlgorithm: 'sha512',
  mgf1HashAlgorithm: 'sha512',
  saltLength: 64,
  publicKeyEncoding: {
    type: 'spki',
    format: 'pem'
  },
  privateKeyEncoding: {
    type: 'pkcs8',
    format: 'pem'
  }
}, callback);
```

### Case 4: ESM import with generateKeyPairSync

**Before:**
```js
import { generateKeyPairSync } from 'node:crypto';

const keys = generateKeyPairSync('rsa-pss', {
  modulusLength: 2048,
  hash: 'sha384',
  mgf1Hash: 'sha384'
});
```

**After:**
```js
import { generateKeyPairSync } from 'node:crypto';

const keys = generateKeyPairSync('rsa-pss', {
  modulusLength: 2048,
  hashAlgorithm: 'sha384',
  mgf1HashAlgorithm: 'sha384'
});
```

### Case 5: Variable options object

**Before:**
```js
const crypto = require('node:crypto');

const options = {
  modulusLength: 2048,
  hash: 'sha256',
  mgf1Hash: 'sha1',
  saltLength: 20
};

crypto.generateKeyPair('rsa-pss', options, (err, publicKey, privateKey) => {
  // handle keys
});
```

**After:**
```js
const crypto = require('node:crypto');

const options = {
  modulusLength: 2048,
  hashAlgorithm: 'sha256',
  mgf1HashAlgorithm: 'sha1',
  saltLength: 20
};

crypto.generateKeyPair('rsa-pss', options, (err, publicKey, privateKey) => {
  // handle keys
});
```

### Case 6: Namespace import

**Before:**
```js
import * as crypto from 'node:crypto';

crypto.generateKeyPair('rsa-pss', {
  modulusLength: 3072,
  hash: 'sha256'
}, callback);
```

**After:**
```js
import * as crypto from 'node:crypto';

crypto.generateKeyPair('rsa-pss', {
  modulusLength: 3072,
  hashAlgorithm: 'sha256'
}, callback);
```

### Case 7: Promise-based usage

**Before:**
```js
const { promisify } = require('node:util');
const crypto = require('node:crypto');

const generateKeyPairAsync = promisify(crypto.generateKeyPair);

generateKeyPairAsync('rsa-pss', {
  modulusLength: 2048,
  hash: 'sha256',
  mgf1Hash: 'sha256'
}).then(({ publicKey, privateKey }) => {
  // handle keys
});
```

**After:**
```js
const { promisify } = require('node:util');
const crypto = require('node:crypto');

const generateKeyPairAsync = promisify(crypto.generateKeyPair);

generateKeyPairAsync('rsa-pss', {
  modulusLength: 2048,
  hashAlgorithm: 'sha256',
  mgf1HashAlgorithm: 'sha256'
}).then(({ publicKey, privateKey }) => {
  // handle keys
});
```

### Case 8: Only hash option present

**Before:**
```js
const crypto = require('node:crypto');

const keyPair = crypto.generateKeyPairSync('rsa-pss', {
  modulusLength: 2048,
  hash: 'sha512'
});
```

**After:**
```js
const crypto = require('node:crypto');

const keyPair = crypto.generateKeyPairSync('rsa-pss', {
  modulusLength: 2048,
  hashAlgorithm: 'sha512'
});
```

### Case 9: Only mgf1Hash option present

**Before:**
```js
const crypto = require('node:crypto');

crypto.generateKeyPair('rsa-pss', {
  modulusLength: 4096,
  mgf1Hash: 'sha384',
  saltLength: 48
}, (err, publicKey, privateKey) => {
  if (err) throw err;
  // use keys
});
```

**After:**
```js
const crypto = require('node:crypto');

crypto.generateKeyPair('rsa-pss', {
  modulusLength: 4096,
  mgf1HashAlgorithm: 'sha384',
  saltLength: 48
}, (err, publicKey, privateKey) => {
  if (err) throw err;
  // use keys
});
```

### Case 10: Should not affect other key types

**Before:**
```js
const crypto = require('node:crypto');

// This should NOT be transformed (not rsa-pss)
crypto.generateKeyPair('rsa', {
  modulusLength: 2048,
  hash: 'sha256'  // This is not the deprecated option for regular RSA
}, callback);

// This SHOULD be transformed (rsa-pss)
crypto.generateKeyPair('rsa-pss', {
  modulusLength: 2048,
  hash: 'sha256'
}, callback);
```

**After:**
```js
const crypto = require('node:crypto');

// This should NOT be transformed (not rsa-pss)
crypto.generateKeyPair('rsa', {
  modulusLength: 2048,
  hash: 'sha256'  // This is not the deprecated option for regular RSA
}, callback);

// This SHOULD be transformed (rsa-pss)
crypto.generateKeyPair('rsa-pss', {
  modulusLength: 2048,
  hashAlgorithm: 'sha256'
}, callback);
```

## Refs

- [DEP0154: RSA-PSS generate key pair options](https://nodejs.org/api/deprecations.html#dep0154-rsa-pss-generate-key-pair-options)
- https://github.com/nodejs/node/pull/58706

nievasdev:
## Summary
  Add codemod to transform deprecated RSA-PSS crypto options:
  - `hash` → `hashAlgorithm`
  - `mgf1Hash` → `mgf1HashAlgorithm`

  ## Changes
  - Only transforms `crypto.generateKeyPair()` and `crypto.generateKeyPairSync()` calls with `'rsa-pss'` key type
  - Supports both direct calls (`crypto.generateKeyPair`) and destructured imports (`generateKeyPair`)
  - Comprehensive test coverage for various usage patterns

  ## Test plan
  - [x] Tests pass for basic transformations
  - [x] Tests pass for destructured imports
  - [x] Tests verify non-rsa-pss calls are ignored
  - [x] Lint and type checks pass

Resolves #144

### DIFF
--- a/recipes/crypto-rsa-pss-update/README.md
+++ b/recipes/crypto-rsa-pss-update/README.md
@@ -1,0 +1,67 @@
+# crypto-rsa-pss-update
+
+Codemod to handle Node.js crypto deprecation DEP0154 by transforming deprecated RSA-PSS key generation options.
+
+## What it does
+
+This codemod transforms deprecated RSA-PSS crypto options in `crypto.generateKeyPair()` and `crypto.generateKeyPairSync()` calls:
+
+- `hash` → `hashAlgorithm`
+- `mgf1Hash` → `mgf1HashAlgorithm`
+
+The transformation only applies to calls with `'rsa-pss'` as the key type.
+
+## Examples
+
+### Before
+
+```javascript
+const crypto = require('crypto');
+
+crypto.generateKeyPair('rsa-pss', {
+  modulusLength: 2048,
+  hash: 'sha256',
+  mgf1Hash: 'sha1',
+  saltLength: 32
+}, (err, publicKey, privateKey) => {
+  // callback
+});
+
+crypto.generateKeyPairSync('rsa-pss', {
+  modulusLength: 2048,
+  hash: 'sha256'
+});
+```
+
+### After
+
+```javascript
+const crypto = require('crypto');
+
+crypto.generateKeyPair('rsa-pss', {
+  modulusLength: 2048,
+  hashAlgorithm: 'sha256',
+  mgf1HashAlgorithm: 'sha1',
+  saltLength: 32
+}, (err, publicKey, privateKey) => {
+  // callback
+});
+
+crypto.generateKeyPairSync('rsa-pss', {
+  modulusLength: 2048,
+  hashAlgorithm: 'sha256'
+});
+```
+
+## Usage
+
+```bash
+npx codemod@next @nodejs/crypto-rsa-pss-update
+```
+
+## Supports
+
+- Both `crypto.generateKeyPair()` and `crypto.generateKeyPairSync()`
+- Destructured imports: `const { generateKeyPair } = require('crypto')`
+- Only transforms `'rsa-pss'` key type calls
+- Preserves all other options and call structure
--- a/recipes/crypto-rsa-pss-update/package.json
+++ b/recipes/crypto-rsa-pss-update/package.json
@@ -12,7 +12,7 @@
     "directory": "recipes/crypto-rsa-pss-update",
     "bugs": "https://github.com/nodejs/userland-migrations/issues"
   },
-  "author": "Claude Code",
+  "author": "Mauro Nievas",
   "license": "MIT",
   "homepage": "https://github.com/nodejs/userland-migrations/blob/main/recipes/crypto-rsa-pss-update/README.md",
   "devDependencies": {
--- a/recipes/crypto-rsa-pss-update/package.json
+++ b/recipes/crypto-rsa-pss-update/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "@nodejs/crypto-rsa-pss-update",
+  "version": "1.0.0",
+  "description": "Handle DEP0154 via transforming deprecated RSA-PSS crypto options `hash` to `hashAlgorithm` and `mgf1Hash` to `mgf1HashAlgorithm`.",
+  "type": "module",
+  "scripts": {
+    "test": "npx codemod@next jssg test -l typescript ./src/workflow.ts ./"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/nodejs/userland-migrations.git",
+    "directory": "recipes/crypto-rsa-pss-update",
+    "bugs": "https://github.com/nodejs/userland-migrations/issues"
+  },
+  "author": "Claude Code",
+  "license": "MIT",
+  "homepage": "https://github.com/nodejs/userland-migrations/blob/main/recipes/crypto-rsa-pss-update/README.md",
+  "devDependencies": {
+    "@types/node": "^24.0.3",
+    "@codemod.com/jssg-types": "^1.0.3"
+  },
+  "dependencies": {
+    "@nodejs/codemod-utils": "*"
+  }
+}
--- a/recipes/crypto-rsa-pss-update/src/workflow.ts
+++ b/recipes/crypto-rsa-pss-update/src/workflow.ts
@@ -1,0 +1,83 @@
+import type { SgRoot, Edit } from "@codemod.com/jssg-types/main";
+
+/**
+ * Transform function that updates deprecated RSA-PSS crypto options.
+ * 
+ * Transforms:
+ * - hash -> hashAlgorithm
+ * - mgf1Hash -> mgf1HashAlgorithm
+ * 
+ * Only applies to crypto.generateKeyPair() and crypto.generateKeyPairSync() 
+ * calls with 'rsa-pss' as the first argument.
+ */
+export default function transform(root: SgRoot): string | null {
+	const rootNode = root.root();
+	let hasChanges = false;
+	const edits: Edit[] = [];
+
+	// Find all generateKeyPair and generateKeyPairSync calls
+	const generateKeyPairCalls = rootNode.findAll({
+		rule: {
+			any: [
+				{ pattern: "crypto.generateKeyPair($TYPE, $OPTIONS, $CALLBACK)" },
+				{ pattern: "crypto.generateKeyPair($TYPE, $OPTIONS)" },
+				{ pattern: "generateKeyPair($TYPE, $OPTIONS, $CALLBACK)" },
+				{ pattern: "generateKeyPair($TYPE, $OPTIONS)" }
+			]
+		}
+	});
+
+	const generateKeyPairSyncCalls = rootNode.findAll({
+		rule: {
+			any: [
+				{ pattern: "crypto.generateKeyPairSync($TYPE, $OPTIONS)" },
+				{ pattern: "generateKeyPairSync($TYPE, $OPTIONS)" }
+			]
+		}
+	});
+
+	// Process both generateKeyPair and generateKeyPairSync calls
+	const allCalls = [...generateKeyPairCalls, ...generateKeyPairSyncCalls];
+
+	for (const call of allCalls) {
+		const typeMatch = call.getMatch("TYPE");
+		const optionsMatch = call.getMatch("OPTIONS");
+
+		if (!typeMatch || !optionsMatch) continue;
+
+		// Only process 'rsa-pss' key type
+		const typeText = typeMatch.text();
+		if (!typeText.includes("'rsa-pss'") && !typeText.includes('"rsa-pss"')) {
+			continue;
+		}
+
+		const optionsText = optionsMatch.text();
+		let newOptionsText = optionsText;
+		let optionsChanged = false;
+
+		// Transform hash -> hashAlgorithm
+		if (optionsText.includes("hash:")) {
+			// Match hash: followed by value, but not hashAlgorithm:
+			newOptionsText = newOptionsText.replace(/\bhash:\s*/g, "hashAlgorithm: ");
+			optionsChanged = true;
+		}
+
+		// Transform mgf1Hash -> mgf1HashAlgorithm
+		if (optionsText.includes("mgf1Hash:")) {
+			newOptionsText = newOptionsText.replace(/\bmgf1Hash:\s*/g, "mgf1HashAlgorithm: ");
+			optionsChanged = true;
+		}
+
+		if (optionsChanged) {
+			// Replace the entire call with updated options
+			const callText = call.text();
+			const newCallText = callText.replace(optionsText, newOptionsText);
+			edits.push(call.replace(newCallText));
+			hasChanges = true;
+		}
+	}
+
+	if (!hasChanges) return null;
+
+	return rootNode.commitEdits(edits);
+}
--- a/recipes/crypto-rsa-pss-update/workflow.yaml
+++ b/recipes/crypto-rsa-pss-update/workflow.yaml
@@ -1,0 +1,25 @@
+version: "1"
+
+nodes:
+  - id: apply-transforms
+    name: Apply AST Transformations
+    type: automatic
+    runtime:
+      type: direct
+    steps:
+      - name: Replace deprecated RSA-PSS crypto options `hash` to `hashAlgorithm` and `mgf1Hash` to `mgf1HashAlgorithm`.
+        js-ast-grep:
+          js_file: src/workflow.ts
+          base_path: .
+          include:
+            - "**/*.js"
+            - "**/*.jsx"
+            - "**/*.mjs"
+            - "**/*.cjs"
+            - "**/*.cts"
+            - "**/*.mts"
+            - "**/*.ts"
+            - "**/*.tsx"
+          exclude:
+            - "**/node_modules/**"
+          language: typescript
--- a/package-lock.json
+++ b/package-lock.json
@@ -1410,6 +1410,10 @@
       "resolved": "recipes/create-require-from-path",
       "link": true
     },
+    "node_modules/@nodejs/crypto-rsa-pss-update": {
+      "resolved": "recipes/crypto-rsa-pss-update",
+      "link": true
+    },
     "node_modules/@nodejs/import-assertions-to-attributes": {
       "resolved": "recipes/import-assertions-to-attributes",
       "link": true
@@ -4063,6 +4067,18 @@
         "@types/node": "^24.0.3"
       }
     },
+    "recipes/crypto-rsa-pss-update": {
+      "name": "@nodejs/crypto-rsa-pss-update",
+      "version": "1.0.0",
+      "license": "MIT",
+      "dependencies": {
+        "@nodejs/codemod-utils": "*"
+      },
+      "devDependencies": {
+        "@codemod.com/jssg-types": "^1.0.3",
+        "@types/node": "^24.0.3"
+      }
+    },
     "recipes/import-assertions-to-attributes": {
       "name": "@nodejs/import-assertions-to-attributes",
       "version": "1.0.0",
@@ -4074,7 +4090,7 @@
     },
     "recipes/process-main-module": {
       "name": "@nodejs/process-main-module",
-      "version": "1.0.0",
+      "version": "1.0.1",
       "license": "MIT",
       "dependencies": {
         "@nodejs/codemod-utils": "0.0.0"
--- a/recipes/crypto-rsa-pss-update/codemod.yaml
+++ b/recipes/crypto-rsa-pss-update/codemod.yaml
@@ -1,0 +1,23 @@
+schema_version: "1.0"
+name: "@nodejs/crypto-rsa-pss-update"
+version: 1.0.0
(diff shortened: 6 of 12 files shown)


recipes/crypto-rsa-pss-update/README.md
```diff
@@ -0,0 +1,67 @@
...
+- `hash` → `hashAlgorithm`
+- `mgf1Hash` → `mgf1HashAlgorithm`
+
+The transformation only applies to calls with `'rsa-pss'` as the key type.
+
+## Examples
+
+### Before
+
+```javascript
```
AugustinMauroy, line 18:
```suggestion
**Before:**
```js
```

recipes/crypto-rsa-pss-update/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+crypto.generateKeyPairSync('rsa-pss', {
+  modulusLength: 2048,
+  hash: 'sha256'
+});
+```
+
+### After
+
+```javascript
```
AugustinMauroy, line 38:
```suggestion
**After:**
```js
```

AugustinMauroy (commented):
good starting ! But there are missing piece:
- if user do `import * as foo from 'node:crypto'`
- if use do `const { generateKeyPair: foo } = require('node:crypto')`
Solution: have this pr merged https://github.com/nodejs/userland-migrations/pull/151

Also for strictness on parameters analyst you should do something like that:
```js
for (const call of callNode) {
	const optionsMatch = call.getMatch("OPTIONS");
	if (!optionsMatch) continue;

	// Find the object node that contains the URL options
	const objectNode = optionsMatch.find({
		rule: {
			kind: "object"
		}
	});

	if (!objectNode) continue;

	// Find all property pairs in the object
	const pairs = objectNode.findAll({
		rule: {
			kind: "pair"
		}
	});

	// find hash pair then replace it
	pairs.find(pair => {
		const keyNode = pair.find({
				rule: {
					kind: "property_identifier"
				}
			});
		if (!keyNode) return false;

		const key = keyNode.text();
		if (key === "hash") {
			const valueNode = pair.find({
				rule: {
					kind: "string_fragment"
				}
			}) || pair.find({
				rule: {
					kind: "string"
				}
			});

			if (!valueNode) return false;

			const value = valueNode.text().slice(1, -1); // Remove quotes
			const replacement = `hashAlgorithm: '${value}'`;
			edits.push(call.replace(replacement));
			return true;
		}
		return false;
	});
}
```

recipes/crypto-rsa-pss-update/tests/expected/import-star-as.js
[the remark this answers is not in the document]
brunocroh: Sorry, but is not clear for me why `resolveBindingPath` do not resolve this scenario, because It can be used to handle different bindings in this way presented below.

```js
`import { generateKeyPair, createHash, randomBytes } from 'crypto';`

const generateKeyPairBind = resolveBindingPath(node, '$.generateKeyPair')
const generateKeyPairSyncBind = resolveBindingPath(node, '$.generateKeyPairSync')

const bindings = [generateKeyPairBind, generateKeyPairSyncBind] //or using Set as well

for(const bind of bindings){
 //handle it searcing by all references to this bind in the file
}
```

The resolveBindingPath, just will return to you the variable name(binding) this function has in the actual file, independent of the way it was imported, require, import, namespace, alias, destructured object.

And one more advantage is ff in the future a new require/import way need to be handled we can just update it in the utility function over go back to each recipe to update it

recipes/crypto-rsa-pss-update/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "type": "git",
+    "url": "git+https://github.com/nodejs/userland-migrations.git",
+    "directory": "recipes/crypto-rsa-pss-update",
+    "bugs": "https://github.com/nodejs/userland-migrations/issues"
+  },
+  "author": "Mauro Nievas",
+  "license": "MIT",
+  "homepage": "https://github.com/nodejs/userland-migrations/blob/main/recipes/crypto-rsa-pss-update/README.md",
+  "devDependencies": {
+    "@types/node": "^24.0.3",
```
AugustinMauroy, line 19:
```suggestion
```

this codemod isn't run on node context

recipes/crypto-rsa-pss-update/src/workflow.ts
```diff
@@ -0,0 +1,160 @@
+import type { SgRoot, SgNode, Edit } from "@codemod.com/jssg-types/main";
+import { getNodeImportStatements } from "@nodejs/codemod-utils/ast-grep/import-statement";
+import { getNodeRequireCalls } from "@nodejs/codemod-utils/ast-grep/require-call";
+import { resolveBindingPath } from "@nodejs/codemod-utils/ast-grep/resolve-binding-path";
+import type JS from "@codemod.com/jssg-types/langs/javascript";
```
AugustinMauroy, line 5:
```suggestion
import type { SgRoot, SgNode, Edit } from "@codemod.com/jssg-types/main";
import type JS from "@codemod.com/jssg-types/langs/javascript";
import { getNodeImportStatements } from "@nodejs/codemod-utils/ast-grep/import-statement";
import { getNodeRequireCalls } from "@nodejs/codemod-utils/ast-grep/require-call";
import { resolveBindingPath } from "@nodejs/codemod-utils/ast-grep/resolve-binding-path";
```

nievasdev: Hi Agustin, Why the change in the order of imports?

AugustinMauroy: it's prettier like that beaucause its:
- package import
- local thing

nievasdev: oh good, thank you!

recipes/crypto-rsa-pss-update/tests/expected/promisified-wrappers.js
```diff
@@ -0,0 +1,46 @@
+const crypto = require('crypto');
+const util = require('util');
+
+// Case 1: Basic promisified wrapper
+const generateKeyPairAsync = util.promisify(crypto.generateKeyPair);
```
brunocroh, line 5: Create other test with variantion of promisify import

```suggestion
const {promisify} = require('util');

// Case 1: Basic promisified wrapper
const generateKeyPairAsync = promisify(crypto.generateKeyPair);
```

recipes/crypto-rsa-pss-update/src/workflow.ts
```diff
@@ -217,9 +219,43 @@ function getCryptoBindings(root: SgRoot<JS>): string[] {
...
+	const rootNode = root.root();
+
+	for (const binding of existingBindings) {
+		// Find: const someVar = util.promisify(crypto.generateKeyPair)
+		// or:   const someVar = util.promisify(generateKeyPair) 
+		const promisified = rootNode.findAll({
+			rule: {
+				pattern: `const $BINDING = util.promisify(${binding})`
+			}
+		});
```
brunocroh, line 243: I think it not will works in scenario below

`const {promisify} = require("node:util")` 

probably you will need use`resolveBindingPath` for promisify

```suggestion
		// or:   const someVar = util.promisify(generateKeyPair) 
		const promisifyBind = resolveBindingPath(node, "$.promisify")
		const promisified = rootNode.findAll({
			rule: {
				pattern: `const $BINDING = ${promisifyBind}(${binding})`
			}
		});
```

brunocroh (approved):
Nice work, thanks! 🚀

nievasdev:
> Nice work, thanks! 🚀

Thank you @brunocroh !

recipes/crypto-rsa-pss-update/workflow.yaml
```diff
@@ -0,0 +1,30 @@
...
+nodes:
+  - id: apply-transforms
+    name: Apply AST Transformations
+    type: automatic
+    runtime:
+      type: direct
+    steps:
+      - name: >-
+          Replace deprecated RSA-PSS crypto options `hash` to `hashAlgorithm`
+          and `mgf1Hash` to `mgf1HashAlgorithm`.
```
brunocroh, line 15: I didn’t understand why this line has >-.
Is it a YAML convention or something to prevent exceeding the column limit?

nievasdev: Based on the following documentation: [official YAML multiline documentation](https://yaml-multiline.info/)

  **Block Scalar syntax has two parts:**
  1. **`>` (folded style)** - "use the folded style, indicated by a right angle bracket (>)" - replaces newlines with spaces
  2. **`-` (strip chomping)** - "To remove all newlines, strip them by putting a minus sign (-) after the style indicator"

  So `>-` combines both: **folded + strip** = multiline for readability but results in a single-line string without trailing newline.

I think is ok,
I should have done it differently?

brunocroh: I didn’t know about these folded styles and strip chomping, thank you for the explanation 😄

nievasdev: Jaja, No problem, Thank you bruno!  😄

recipes/crypto-rsa-pss-update/README.md
```diff
@@ -0,0 +1,75 @@
...
+crypto.generateKeyPairSync("rsa-pss", {
+	modulusLength: 2048,
+	hashAlgorithm: "sha256",
+});
+```
+
+## Usage
+
+```bash
+npx codemod@next @nodejs/crypto-rsa-pss-update
```
AugustinMauroy, line 67:
```suggestion
npx codemod @nodejs/crypto-rsa-pss-update
```

no needed anymore

recipes/crypto-rsa-pss-update/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@nodejs/crypto-rsa-pss-update",
+  "version": "1.0.0",
+  "description": "Handle DEP0154 via transforming deprecated RSA-PSS crypto options `hash` to `hashAlgorithm` and `mgf1Hash` to `mgf1HashAlgorithm`.",
+  "type": "module",
+  "scripts": {
+    "test": "npx codemod@next jssg test -l typescript ./src/workflow.ts ./"
```
AugustinMauroy, line 7:
```suggestion
    "test": "npx codemod jssg test -l typescript ./src/workflow.ts ./"
```

ditto

recipes/crypto-rsa-pss-update/src/workflow.ts
```diff
@@ -0,0 +1,310 @@
...
+		// Check if it's a string literal 'rsa-pss' or "rsa-pss"
+		const isStringLiteral = /^['"]rsa-pss['"]$/.test(typeText);
+		
+		// Check if it's a variable that resolves to 'rsa-pss'
+		const isVariableWithRsaPss = /^[a-zA-Z_$][a-zA-Z0-9_$]*$/.test(typeText) && 
+			resolveVariableValue(rootNode, typeText);
+
+		if (!isStringLiteral && !isVariableWithRsaPss) {
+			continue;
+		}
```
AugustinMauroy, line 109:
```suggestion
		if (!isStringLiteral && !isVariableWithRsaPss) continue;
```

recipes/crypto-rsa-pss-update/src/workflow.ts
```diff
@@ -0,0 +1,310 @@
...
+					}
+				}) || pair.find({
+					rule: {
+						kind: "binary_expression"  // For cases like 'sha' + '256'
+					}
+				}) || pair.find({
+					rule: {
+						kind: "ternary_expression"  // For cases like condition ? 'sha512' : 'sha256'
+					}
+				});
```
AugustinMauroy, line 176: Can we use any rule https://ast-grep.github.io/guide/rule-config/composite-rule.html#any

nievasdev:
@brunocroh @AugustinMauroy. The code was getting big, so I refactored to a modular architecture.

brunocroh:
> @brunocroh @AugustinMauroy. The code was getting big, so I refactored to a modular architecture.

IMHO: The previous code probably had some areas that could be improved, but it was easier to follow.

> TLDR: workflow.ts was split, but I don’t see much benefit. Splitting makes sense only when a file has multiple responsibilities or needs independent changes (like PR #172). Adding extra files here hides logic and makes it harder to follow. A consistent structure across codemods would help, but one-off architecture/splits just add complexity.

I can see that workflow.ts was split, but I don’t see any real advantage to this. Splits like this make sense when a file has issues with [Separation of Concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) or when we’re trying to follow the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

In our workflow, we have a script that rewrites the file. I believe splitting the file only makes sense in cases where different parts need to change independently, like what PR #172 shows. In those scenarios, having a shared file for reusable parts can make sense.

In many other cases, adding extra files and utility functions doesn’t simplify the code. Instead, it hides what’s happening, making the logic harder to follow when it’s split across 3 or 4 files.

Maybe in the future, we can create a structure or architecture to apply consistently across all codemod recipes, making them easier to follow. However, the real advantage of such structures comes when they form a pattern used across the project. Having just one codemod recipe with a different and more complicated pattern will only make it harder to understand, as it will be completely different from the rest.

---

Again, just my opinion. Let’s see what the maintainers think.

AugustinMauroy:
> Again, just my opinion. Let’s see what the maintainers think.

Agreed with Bruno. But @nievasdev I understand that you spend a lot of time on your PR and that it takes a lot of energy. And I'm very happy to have people who are so invested in the project.

nievasdev:
@brunocroh you got a point, I think you're right, I'll make the code structure like the previous ones,
@AugustinMauroy  you spend a lot of time too, 
I'm not going to abandon the project. I understand that these first issues are for learning, so don't worry about me getting discouraged, haha.
And thank you very much for taking the time to review my commits.

brunocroh:
> @brunocroh @AugustinMauroy, how are you? Are there other things we should change? How should I continue with this work?

We need one more member to review/approve before proceeding with the merge. 
We are currently waiting for feedback from the team.

AugustinMauroy:
> Good, could I take another issue while I wait, or is it better if I just wait?

yeah if you want you can open an second pr

PR #165, opened 2025-08-07
fix(create-require-from-path): handle require/import with alias

### LINKED ISSUE #142: feat(`create-require-from-path`): import alias isn't handled
## Actual behavior on `v1.0.0`

**Before**
```js
import { createRequireFromPath as foo } from '../index.mjs';

const r3 = foo('/path/to/module');
```

**After:**
```js
import { createRequireFromPath as foo } from '../index.mjs';

var r1 = foo('/path/to/module');
let r2 = foo('/path/to/module');
const r3 = foo('/path/to/module');
```

## Expected behavior

**Before**
```js
import { createRequireFromPath as foo } from '../index.mjs';

const r3 = foo('/path/to/module');
```

**After:**
```js
import { createRequire as foo } from '../index.mjs';

var r1 = foo('/path/to/module');
let r2 = foo('/path/to/module');
const r3 = foo('/path/to/module');
```

brunocroh:
closes #142

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4053,7 +4053,7 @@
     },
     "recipes/create-require-from-path": {
       "name": "@nodejs/create-require-from-path",
-      "version": "1.0.0",
+      "version": "1.1.0",
       "license": "MIT",
       "dependencies": {
         "@nodejs/codemod-utils": "0.0.0"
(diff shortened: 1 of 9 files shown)


brunocroh:
> BTW need to bump the minor so we can release it. it's a minor because we add new "features" (I mean supported cases )

Brilliant, done then

AugustinMauroy:
> > BTW need to bump the minor so we can release it. it's a minor because we add new "features" (I mean supported cases )
> 
> Brilliant, done then

`npm i` to update `package-lock.json` 😁

AugustinMauroy (approved):

Issue #92, opened 2025-06-24
Brainstorm: `url.parse()` -> `new url`


AugustinMauroy:
## Description

This codemod should migrate `url.parse()` to `new url`. 

- It's should handle both `require()` function and `import` statement.
- It's should support both `url` `node:url` specifier.
- If the url module isn't use elsewhere we should remove the importation.
- `auth` property should be replaced by `${modernURL.username}:${modernURL.password}`
- `path` property should be replaced by `${modernURL.pathname}${modernURL.search}`
- `hostname` property should be replaced by `modernURL.hostname.replace(/^\[|\]$/, '')` to remove the brackets for IPv6 addresses.

## Example

> In these example we use `legacyURL` to refer to the object returned by `url.parse()`, and `modernURL` to refer to the object created with `new URL()`.
> In the codemod implementation, it' will be the name of the variable that is used by the user so it's shouldn't be changed.

### Case 1: Basic usage

**Before:**
```js
const url = require('node:url');

const legacyURL = url.parse('https://example.com/path?query=string#hash');
```

**After:**
```js
const modernURL = new URL('https://example.com/path?query=string#hash');
```

### Case 2: `auth` property

**Before:**
```js
const url = require('node:url');

const legacyURL = url.parse('https://user:pass@example.com/path');

const { auth } = legacyURL;
const urlAuth = legacyURL.auth;
```

**After:**
```js
const modernURL = new URL('https://user:pass@example.com/path');

const auth  = `${modernURL.username}:${modernURL.password}`;
const urlAuth = `${modernURL.username}:${modernURL.password}`;
```

### Case 3: `path` property

**Before:**
```js
const url = require('node:url');

const legacyURL = url.parse('https://example.com/path?query=string#hash');

const { path } = legacyURL;
const urlPath = legacyURL.path;
```

**After:**
```js
const modernURL = new URL('https://example.com/path?query=string#hash');

const path = `$(modernURL.pathname}${modernURL.search}`;
const urlPath = `${modernURL.pathname}${modernURL.search}`;
```

### Case 4: `hostname` property with IPv6 address

**Before:**
```js
const url = require('node:url');

const legacyURL = url.parse('https://[2001:db8::1]/path');

const { hostname } = legacyURL;
const urlHostname = legacyURL.hostname;
```

**After:**
```js
const modernURL = new URL('https://[2001:db8::1]/path');

const hostname = modernURL.hostname.replace(/^\[|\]$/, '');
[...]

PR #172, opened 2025-08-12
Feat(`node url to whatwg url`)

AugustinMauroy:
## Description

Add codemod for `node:url` legacy object to whatWG url

## Related issue 

close https://github.com/nodejs/userland-migrations/issues/92

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1497,6 +1497,10 @@
       "resolved": "recipes/process-main-module",
       "link": true
     },
+    "node_modules/@nodejs/repl-builtin-modules": {
+      "resolved": "recipes/repl-builtin-modules",
+      "link": true
+    },
     "node_modules/@nodejs/rmdir": {
       "resolved": "recipes/rmdir",
       "link": true
@@ -4344,6 +4348,17 @@
       },
       "devDependencies": {
         "@codemod.com/jssg-types": "^1.0.9"
+      }
+    },
+    "recipes/repl-builtin-modules": {
+      "name": "@nodejs/repl-builtin-modules",
+      "version": "1.0.0",
+      "license": "MIT",
+      "dependencies": {
+        "@nodejs/codemod-utils": "*"
+      },
+      "devDependencies": {
+        "@codemod.com/jssg-types": "^1.0.3"
       }
     },
     "recipes/rmdir": {
--- a/recipes/repl-builtin-modules/README.md
+++ b/recipes/repl-builtin-modules/README.md
@@ -1,0 +1,23 @@
+# `repl.builtinModules` DEP0191
+
+This recipe provides a guide for migrating from the deprecated `repl.builtinModules` & `repl._builtinLibs` to the new `module.builtinModules` property in Node.js.
+
+See [DEP0191](https://nodejs.org/api/deprecations.html#DEP0191) and [DEP0142](https://nodejs.org/api/deprecations.html#DEP0142)
+
+## Examples
+
+**Before:**
+```js
+import repl from 'node:repl';
+
+console.log(repl.builtinModules);
+console.log(repl._builtinLibs);
+```
+
+**After:**
+```js
+import module from 'node:module';
+
+console.log(module.builtinModules);
+console.log(module.builtinModules);
+```
--- a/recipes/repl-builtin-modules/codemod.yaml
+++ b/recipes/repl-builtin-modules/codemod.yaml
@@ -1,0 +1,22 @@
+schema_version: "1.0"
+name: "@nodejs/repl-builtin-modules"
+version: 1.0.0
+description: Handle DEP0191 via transforming `repl.builtinModules` usage to `module.builtinModules`.
+author: Augustin Mauroy
+license: MIT
+workflow: workflow.yaml
+category: migration
+
+targets:
+  languages:
+    - javascript
+    - typescript
+
+keywords:
+  - transformation
+  - migration
+  - deprecation
+
+registry:
+  access: public
+  visibility: public
--- a/recipes/repl-builtin-modules/package.json
+++ b/recipes/repl-builtin-modules/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "@nodejs/repl-builtin-modules",
+  "version": "1.0.0",
+  "description": "Handle DEP0191 via transforming `repl.builtinModules` usage to `module.builtinModules`.",
+  "type": "module",
+  "scripts": {
+    "test": "npx codemod@next jssg test -l typescript ./src/workflow.ts ./",
+    "test:update-snapshots": "npx codemod@next jssg test --update-snapshots -l typescript ./src/workflow.ts ./ "
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/nodejs/userland-migrations.git",
+    "directory": "recipes/repl-builtin-modules",
+    "bugs": "https://github.com/nodejs/userland-migrations/issues"
+  },
+  "author": "Augustin Mauroy",
+  "license": "MIT",
+  "homepage": "https://github.com/nodejs/userland-migrations/blob/main/recipes/repl-builtin-modules/README.md",
+  "devDependencies": {
+    "@codemod.com/jssg-types": "^1.0.3"
+  },
+  "dependencies": {
+    "@nodejs/codemod-utils": "*"
+  }
+}
--- a/recipes/repl-builtin-modules/src/workflow.ts
+++ b/recipes/repl-builtin-modules/src/workflow.ts
@@ -1,0 +1,452 @@
+import { EOL } from "node:os";
+import { getNodeImportCalls, getNodeImportStatements } from "@nodejs/codemod-utils/ast-grep/import-statement";
+import { getNodeRequireCalls } from "@nodejs/codemod-utils/ast-grep/require-call";
+import { resolveBindingPath } from "@nodejs/codemod-utils/ast-grep/resolve-binding-path";
+import type { SgRoot, Edit, SgNode } from "@codemod.com/jssg-types/main";
+import type JS from "@codemod.com/jssg-types/langs/javascript";
+
+/**
+ * Regex to match deprecated properties
+ *
+ * _builtinLibs and builtinModules
+ */
+const builtinRegex = /\b(_builtinLibs|builtinModules)\b/g;
+
+// Reusable quoted module specifier for generating new statements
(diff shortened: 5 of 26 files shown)


recipes/node-url-to-whatwg-url/src/url-parse.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    // Destructuring: const { hostname } = obj -> const hostname = obj.hostname.replace(/^\[|\]$/, '')
+    const hostnameDestructures = rootNode.findAll({ rule: { pattern: "const { hostname } = $OBJ" } });
+    for (const node of hostnameDestructures) {
+        const base = node.getMatch("OBJ");
+        if (!base) continue;
+        const hadSemi = /;\s*$/.test(node.text());
+        const replacement = `const hostname = ${base.text()}.hostname.replace(/^\\[|\\]$/, '')${hadSemi ? ';' : ''}`;
+        edits.push(node.replace(replacement));
+    }
```
brunocroh, line 123: I think the logic here can be reutilized between the properties.



When more than one property is destructured at the same time, this behaviour not will work, 

i.g:
`const { auth, path, hostname } = myURL`


something like

```js
const fieldsToReplace => [
  {
    key: 'auth',
    replaceFn: (base, hadSemi) => (`const auth = \`${'${'}${name}.username${'}'}:${'${'}${name}.password${'}'}\`${hadSemi ? ';' : ''}`)
  },
	{
		key: 'path',
    replaceFn: (base, hadSemi) => (const path = \`${'${'}${name}.pathname${'}'}${'${'}${name}.search${'}'}\`${hadSemi ? ';' : ''}`)
	}

]
```

.github/workflows/codemod_publish.yml
```diff
@@ -106,11 +106,11 @@ jobs:
       - name: Authenticate with Codemod registry
         env:
           CODEMOD_TOKEN: ${{ secrets.CODEMOD_TOKEN }}
-        run: npx codemod@next login --api-key "$CODEMOD_TOKEN"
+        run: npx codemod login --api-key "$CODEMOD_TOKEN"
 
       - name: Publish codemod
         working-directory: ${{ steps.parse-tag.outputs.codemod-path }}
-        run: npx codemod@next publish
+        run: npx codemod publish
```
avivkeller, line 113: Can you do this in a separate PR?

recipes/node-url-to-whatwg-url/src/url-parse.ts
```diff
@@ -0,0 +1,126 @@
...
+        const replacement = `\`\${${base.text()}.username}:\${${base.text()}.password}\``;
+        edits.push(node.replace(replacement));
+    }
+
+    // Destructuring: const { auth } = obj -> const auth = `${obj.username}:${obj.password}`
+	const fieldsToReplace = [
+		{
+			key: "auth",
+			replaceFn: (base: string, hadSemi: boolean) =>
+				`const auth = \`\${${base}.username}:\${${base}.password}\`${hadSemi ? ";" : ""}`,
```
avivkeller, line 78: Minor minor nit: always using a semicolon is best practice

recipes/node-url-to-whatwg-url/src/url-parse.ts
```diff
@@ -0,0 +1,126 @@
...
+
+            const replacement = `new URL(${arg.text()})`;
+            edits.push(call.replace(replacement));
+        }
+    }
+
+    // 2) Transform legacy properties on URL object
+    //    - auth => `${obj.username}:${obj.password}`
+    //    - path => `${obj.pathname}${obj.search}`
+    //    - hostname => obj.hostname.replace(/^[\[|\]]$/, '')  (strip square brackets)
```
avivkeller, line 61: Does the host name have square brackets?

.github/workflows/codemod_publish.yml
AugustinMauroy: IDK what happened this change is already merged on main. maybe an issue on my rebase

recipes/node-url-to-whatwg-url/src/url-parse.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    // Property access: obj.auth -> `${obj.username}:${obj.password}`
+    const authAccesses = rootNode.findAll({ rule: { pattern: "$OBJ.auth" } });
```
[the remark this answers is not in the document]
AugustinMauroy: > I think it would be good to add a test that covers this scenario as well.

Agreed with you buddy

brunocroh (approved):

recipes/node-url-to-whatwg-url/tests/url-parse/input/file-1.js
```diff
@@ -0,0 +1,18 @@
+const url = require('node:url');
+
+const myURL = url.parse('https://example.com/path?query=string#hash');
```
styfle, line 3: Please add a test for parsing a relative url, for example:

```js
url.parse('/path?query=string#hash')
```

recipes/node-url-to-whatwg-url/src/url-parse.ts
```diff
@@ -0,0 +1,226 @@
...
+		for (const m of matches) {
+			const obj = m.getMatch("OBJ");
+			if (!obj) continue;
+			const name = obj.text();
+			if (/^[$A-Z_a-z][$\w]*$/.test(name)) parseResultVars.add(name);
+		}
+	}
+
+	// 1.b) Replace parse calls with new URL()
+	//      Also, for declarations using `var`, upgrade to `let` while keeping `const` as-is.
```
JakobJingleheimer, line 63: I think that upgrade from `var` → `let` is not safe: `var`'s scoping is different than `let`

```js
function foo(a) {
  switch(a) {
    case '…': {
      var u = url.parse(…);
      let l = u;
    }
  }

  u.slice(0,10); // works
  l.slice(0,10); // 💥
}
```

recipes/node-url-to-whatwg-url/src/url-parse.ts
```diff
@@ -0,0 +1,226 @@
...
+	for (const pattern of parseCallPatterns) {
+		const calls = rootNode.findAll({ rule: { pattern } });
+
+		for (const call of calls) {
+			const arg = call.getMatch("ARG");
+			if (!arg) continue;
+
+			const replacement = `new URL(${arg.text()})`;
+			edits.push(call.replace(replacement));
```
JakobJingleheimer, line 72: nit

```suggestion
			edits.push(call.replace(`new URL(${arg.text()})`));
```

recipes/node-url-to-whatwg-url/src/url-parse.ts
```diff
@@ -0,0 +1,226 @@
...
+    // 2) Transform legacy properties on URL object
+    //    - auth => `${obj.username}:${obj.password}`
+    //    - path => `${obj.pathname}${obj.search}`
+    //    - hostname => obj.hostname.replace(/^[\[|\]]$/, '')  (strip square brackets)
+	const fieldsToReplace = [
+		{
+			key: "auth",
+			replaceFn: (base: string, hadSemi: boolean, declKind: "const" | "let" | "var") => {
+				const kind = declKind === "var" ? "let" : declKind;
+				return `${kind} auth = \`\${${base}.username}:\${${base}.password}\`${hadSemi ? ";" : ""}`;
```
JakobJingleheimer, line 93: If `String.raw()` works/helps elsewhere, I think apply it here too.

JakobJingleheimer (changes requested):
Generally looks good! A couple things though:

* Dynamic imports?
* I think the `var` → `let` needs to get removed because it can easily break user code. (Requesting changes on this)

Also, I think [Steven's request](https://github.com/nodejs/userland-migrations/pull/172/files?short_path=088dc73#r2322115260) is a good one.

nit: I think indentation might be inconsistent across files? (spaces vs tabs)

PS Sorry it took me a bit to get to this 😓

AugustinMauroy:
update:
- add test
- I didn't replace `return undefined` by `return;` because I prefer to be explicit 
- `String.raw` isn't really useful because it's only escape `\` and not `` ` ``

JakobJingleheimer:
@AugustinMauroy is there a deprecation for this?

recipes/node-url-to-whatwg-url/README.md
```diff
@@ -60,6 +60,9 @@ url.format({
 const myUrl = new URL('https://example.com/some/path?page=1&format=json').toString();
 ```
 
+> **Note:** The migration of `url.format` can also be done as `` `${new URL('https://example.com/some/path?page=1&format=json')}` `` which is little bit efficienter.
+> But it's may be less readable for some users.
```
JakobJingleheimer:
```suggestion
> **Note:** The migration of `url.format` can also be done as `` `${new URL('https://example.com/some/path?page=1&format=json')}` `` which is little bit more efficient. But it may be less readable for some users.
```

recipes/node-url-to-whatwg-url/src/url-parse.ts
```diff
@@ -51,11 +53,14 @@ export default function transform(root: SgRoot): string | null {
 				]
 			}
 		});
+
+		const validVariableNameRegex = /^[$A-Z_a-z][$\w]*$/;
```
JakobJingleheimer: nit: I would move this to the root level

brunocroh (approved):

recipes/node-url-to-whatwg-url/tests/url-parse/expected/file-10.js
```diff
@@ -0,0 +1,13 @@
+const url = require('node:url');
+
+const myURL = new URL('/path?query=string#hash');
```
styfle: This throws an error

```
Uncaught TypeError: Failed to construct 'URL': Invalid URL
```

styfle: @AugustinMauroy This codemod is very dangerous because it will cause apps to crash at runtime.

AugustinMauroy: @styfle if you can provide a better before/after example that can be used as test that will help us to fix this issue

JakobJingleheimer:
@AugustinMauroy FYI: there were some undocumented pieces to a recent related deprecation (DEP0169) that may affect this: https://github.com/nodejs/node/pull/61644

PR #176, opened 2025-08-12
chore: remove `@next` and clean

AugustinMauroy:
- remove `@next`
- remove wrong typing to use `jssg` type
- update naming based on codemod team recommendation

### DIFF
--- a/recipes/tmpdir-to-tmpdir/src/workflow.ts
+++ b/recipes/tmpdir-to-tmpdir/src/workflow.ts
@@ -1,6 +1,6 @@
 import { getNodeImportStatements } from "@nodejs/codemod-utils/ast-grep/import-statement";
 import { getNodeRequireCalls } from "@nodejs/codemod-utils/ast-grep/require-call";
-import type { SgRoot, Edit } from "@ast-grep/napi";
+import type { SgRoot, Edit } from "@codemod.com/jssg-types/main";
 
 /**
  * Transform function that updates code to replace deprecated `tmpDir` usage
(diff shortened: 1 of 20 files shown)


brunocroh (approved):

recipes/tmpdir-to-tmpdir/src/workflow.ts
```diff
@@ -26,6 +26,7 @@ export default function transform(root: SgRoot): string | null {
   let hasChanges = false;
 
   // Step 1: Find and update destructuring assignments from require('os') or require('node:os')
+	// @ts-ignore - ast-grep types are not fully compatible with JSSG types
```
JakobJingleheimer, line 29: I think better to use `@ts-expect-error` so it alerts us when the issue is fixed.

recipes/tmpdir-to-tmpdir/src/workflow.ts
```diff
@@ -47,6 +48,7 @@ export default function transform(root: SgRoot): string | null {
     }
   }
 
+	// @ts-ignore - ast-grep types are not fully compatible with JSSG types
```
JakobJingleheimer, line 51: Same: `@ts-expect-error` so it alerts us when the issue is fixed.

JakobJingleheimer (approved):

alexbit-codemod (approved):
this looks good to me. just fyi, @mohebifar previously suggested we leave the @latest tag in the npx commands to make sure users get the latest version.

AugustinMauroy:
> this looks good to me. just fyi, @mohebifar previously suggested we leave the `@latest` tag in the npx commands to make sure users get the latest version.

by default npx use `@lastest` (checked localy) so I prefer to have something shorter

JakobJingleheimer (approved):

PR #177, opened 2025-08-13
feat(crypto-fips): add migration recipe for crypto.fips

### LINKED ISSUE #175: feat: DEP0093: crypto.fips is deprecated and replaced
## Description

This codemod should migrate crypto.fips property usage to crypto.getFips() and crypto.setFips(). It's useful to migrate code that uses the deprecated `crypto.fips` property which has been replaced with dedicated functions.

It should replace property reads with `crypto.getFips()` calls. It should replace property assignments with `crypto.setFips()` calls. It should handle both CommonJS and ESM imports. It should preserve FIPS mode configuration logic.

### Case 1

**Before:**
```js
const crypto = require("node:crypto");
if (crypto.fips) {
  console.log("FIPS mode is enabled");
}
```

**After:**
```js
const crypto = require("node:crypto");
if (crypto.getFips()) {
  console.log("FIPS mode is enabled");
}
```

### Case 2

**Before:**
```js
const crypto = require("node:crypto");
crypto.fips = true;
```

**After:**
```js
const crypto = require("node:crypto");
crypto.setFips(true);
```

### Case 3

**Before:**
```js
const crypto = require("node:crypto");
if (process.env.ENABLE_FIPS === "true") {
  crypto.fips = true;
}
console.log("FIPS enabled:", crypto.fips);
```

**After:**
```js
const crypto = require("node:crypto");
if (process.env.ENABLE_FIPS === "true") {
  crypto.setFips(true);
}
console.log("FIPS enabled:", crypto.getFips());
```

### Case 4

**Before:**
```js
import crypto from "node:crypto";
const fipsStatus = crypto.fips;
crypto.fips = !fipsStatus;
```

**After:**
```js
import crypto from "node:crypto";
const fipsStatus = crypto.getFips();
crypto.setFips(!fipsStatus);
```

## REFS

- [DEP0093: `crypto.fips`](https://nodejs.org/api/deprecations.html#DEP0093)

max-programming:
Closes #175

### DIFF
--- a/recipes/crypto-fips-to-getFips/src/workflow.ts
+++ b/recipes/crypto-fips-to-getFips/src/workflow.ts
@@ -109,10 +109,7 @@ function transformNamespaceUsage(rootNode: SgNode<Js>, base: string): Edit[] {
 		const valueNode = assignment.getMatch("VALUE");
 		if (valueNode) {
 			let value = valueNode.text();
-			value = value.replace(
-				new RegExp(`\\b${escapeRegExp(base)}\\.fips\\b`, "g"),
-				`${base}.getFips()`,
-			);
+			value = value.replace(new RegExp(`\\b${base}\\.fips\\b`, "g"), `${base}.getFips()`);
 			edits.push(assignment.replace(`${base}.setFips(${value})`));
 		}
 	}
@@ -155,7 +152,7 @@ function transformDestructuredUsage(rootNode: SgNode<Js>, binding: string): Edit
 		const valueNode = assignment.getMatch("VALUE");
 		if (valueNode) {
 			let value = valueNode.text();
-			value = value.replace(new RegExp(`\\b${escapeRegExp(binding)}\\b`, "g"), "getFips()");
+			value = value.replace(new RegExp(`\\b${binding}\\b`, "g"), "getFips()");
 			edits.push(assignment.replace(`setFips(${value})`));
 		}
 	}
@@ -190,11 +187,4 @@ function transformDestructuredUsage(rootNode: SgNode<Js>, binding: string): Edit
 	}
 
 	return edits;
-}
-
-/**
- * Escape regexp special characters
- */
-function escapeRegExp(input: string): string {
-	return input.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
 }


max-programming:
Let me know if the `settings.json` change feels unnecessary, if that's the case, I'll remove it

Personally, I think having that helps

avivkeller:
> Let me know if the `settings.json` change feels unnecessary, if that's the case, I'll remove it
> 
> 
> 
> Personally, I think having that helps

You don't have to remove it entirely, but can it be a separate PR (to keep our commits organized)?

Thank you!

recipes/crypto-fips/src/workflow.ts
```diff
@@ -0,0 +1,99 @@
...
+			);
+			edits.push(assign.replace(`${base}.setFips(${transformedValue})`));
+			hasChanges = true;
+		}
+	}
+
+	return { edits, hasChanges };
+}
+
+function replaceReads(rootNode: SgNode, cryptoBases: Set<string>) {
```
AugustinMauroy, line 81: Also here you can add JSdoc

recipes/crypto-fips/src/workflow.ts
```diff
@@ -0,0 +1,99 @@
...
+	// Handle crypto.fips (value-access)
+	const readResult = replaceReads(rootNode, cryptoBases);
+	edits.push(...readResult.edits);
+	hasChanges = readResult.hasChanges;
+
+	if (!hasChanges) return null;
+	return rootNode.commitEdits(edits);
+}
+
+function* getCryptoFipsBasePaths(statements: SgNode[]) {
```
AugustinMauroy, line 39: also here JSdoc

recipes/crypto-fips/src/workflow.ts
```diff
@@ -0,0 +1,99 @@
...
+	for (const stmt of statements) {
+		const resolvedPath = resolveBindingPath(stmt, '$.fips');
+		if (!resolvedPath || !resolvedPath.includes('.')) continue;
+		yield resolvedPath.slice(0, resolvedPath.lastIndexOf('.'));
+	}
+}
+
+function* getAllCryptoFipsBasePaths(root: SgRoot) {
```
AugustinMauroy, line 47: also here JSdoc

recipes/crypto-fips/src/workflow.ts
AugustinMauroy, line 1: Also you can replace the `hasChanges` by `edit.lenght()`

recipes/crypto-fips/README.md
```diff
@@ -0,0 +1,33 @@
+# `crypto.fips` DEP0093
+
+This recipe provides a guide for migrating from the deprecated `crypto.fips` to `crypto.getFips()` and `crypto.setFips()`.
+
+See [DEP0093](https://nodejs.org/api/deprecations.html#DEP0093).
+
+## Examples
+
```
AugustinMauroy, line 8: add `import crypto form 'node:cryptor';` to have real world example

AugustinMauroy (approved):
WOW 😵‍💫 that's an awesome first contribution. I love your usage of `yield`.

Thanks you for your contribution, I hope that you had enjoyed it. If you have any feedback feel it

recipes/crypto-fips/src/workflow.ts
```diff
@@ -0,0 +1,276 @@
...
+			edits.push(...replaceNamedAssignments(rootNode, varName));
+			edits.push(...replaceNamedReads(rootNode, varName));
+		}
+	}
+
+	edits.push(...updateCryptoImportSpecifiers(root));
+	edits.push(...updateCryptoRequireDestructuring(root));
+
+	if (edits.length === 0) return null;
+	return rootNode.commitEdits(edits);
```
AugustinMauroy, line 38:
```suggestion

	return rootNode.commitEdits(edits);
```

max-programming: Ah ok I see now. It's an extra line

AugustinMauroy (approved):

AugustinMauroy:
CI is red here

AugustinMauroy:
No problem !

AugustinMauroy (approved):

recipes/crypto-fips/src/workflow.ts
```diff
@@ -0,0 +1,277 @@
...
+		// Ensure both getFips and setFips are present
+		if (!hasGet) keepTexts.push('getFips');
+		if (!hasSet) keepTexts.push('setFips');
+
+		// Replace the whole `{ ... }`
+		edits.push(namedImports.replace(`{ ${keepTexts.join(', ')} }`));
+	}
+
+	return edits;
+}
```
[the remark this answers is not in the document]
brunocroh: The current updateBinding covers more edge cases than your implementation, and it’s a very common task in other codemods. I think having this centralized in a shared library is very important for the healthy growth of the project.

Let’s add support for many-to-many update imports to our shared library and use it here as well.

brunocroh (changes requested):
We’ll use `updateBinding`, which already exists in our utility library, instead of recreating it here.

So, we need to add support for `updateBinding` to add multiple bindings (and maybe remove as well) at the same time.

recipes/crypto-fips/src/workflow.ts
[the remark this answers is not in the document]
max-programming: @brunocroh Implemented this into the updateBinding function. 

Let me know if it looks good or not (It works but let me know if the code quality is good or it smells)

brunocroh (approved):

JakobJingleheimer:
@max-programming thanks for this and getting it updated! I'll try to get to the review tomorrow.

recipes/crypto-fips/src/workflow.ts
```diff
@@ -0,0 +1,202 @@
...
+				node,
+			});
+		}
+	}
+
+	return bindings;
+}
+
+/**
+ * Transform namespace usage: crypto.fips -> crypto.getFips(), crypto.fips = val -> crypto.setFips(val)
```
JakobJingleheimer: nit

```suggestion
 * Transform namespace usage: crypto.fips → crypto.getFips(), crypto.fips = val → crypto.setFips(val)
```

recipes/crypto-fips/src/workflow.ts
```diff
@@ -0,0 +1,202 @@
...
+
+	for (const read of reads) {
+		edits.push(read.replace(`${base}.getFips()`));
+	}
+
+	return edits;
+}
+
+/**
+ * Transform destructured usage: fips -> getFips(), fips = val -> setFips(val)
```
JakobJingleheimer:
```suggestion
 * Transform destructured usage: fips → getFips(), fips = val → setFips(val)
```

JakobJingleheimer (changes requested):
This looks pretty good! It needs to support dynamic imports—we have utilities to help with that (I referenced them in some code suggestions) and tests for them; I think that will be pretty quick/simple to include. Good to go IMO once those are in 🙂

brunocroh (approved):

.vscode/settings.json
```diff
@@ -8,5 +8,8 @@
     "[markdown]": {
         "editor.wordWrap": "off"
     },
+    "[typescript][javascript]": {
+        "editor.defaultFormatter": "biomejs.biome"
+    },
```
JakobJingleheimer: 👍

PR #182, opened 2025-08-14
feat(util/update-binding): add new utility to update binding in the file

brunocroh:
# Add new updateBinding utility function and make removeBinding a wrapper for it

## Summary
This PR create the new `updateBinding` function. That can be used to update bindings in the file 

##Examples

### Updating a named require/import
`updateBinding(node, 'types', {newBinding: 'newTypes'}`

**Before:**

```js
const { types, diff } = require('node:util');
import { types, diff } = from 'node:util';

```

**After:**

```js
const { newTypes, diff } = require('node:util');
import { newTypes, diff } = from 'node:util';
```
---

### Removing a named require/import
`updateBinding(node, 'types', {newBinding: undefined}`

**Before:**

```js
const { types, diff } = require('node:util');
import { types, diff } = from 'node:util';

```

**After:**

```js
const { diff } = require('node:util');
import { diff } = from 'node:util';
```

---

### Updating a named import with alias
`updateBinding(node, 'renamedTypes', {newBinding: 'newTypes'}`

**Before:**

```js
import { types as renamedTypes, diff } = from 'node:util';

```

**After:**

```js
import { newTypes as renamedTypes, diff } = from 'node:util';
```

## Changes Made

- Created `updateBinding` function in `/utils/src/ast-grep/update-binding.ts`
- Refactored existing `remove-binding` function to use the new `updateBinding` internally


AugustinMauroy (approved):
LGTM ! thanks for that

JakobJingleheimer:
@brunocroh could you please undo the formatting changes? It's hugely bloating the recent diff

utils/src/ast-grep/update-binding.ts
```diff
@@ -0,0 +1,359 @@
...
+ * // And binding: "isNativeError"
+ * // Returns: undefined (no destructured binding found)
+ * ```
+ */
+export function updateBinding(
+	node: SgNode<Js> | SgNode<Js, Kinds<Js>>,
+	options?: UpdateBindingOptions,
+): UpdateBindingReturnType {
+	const nodeKind = node.kind().toString();
+
```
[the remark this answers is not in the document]
AugustinMauroy: what if invalid node is pass as arg it's an error from dev so if we throws a message it's will allow us to have better DX.
also if there are any change we should return null so we know everything work but nothing happened

utils/README.md
```diff
@@ -230,5 +250,3 @@ const { types: t } = require('util'); // → t.isNativeError
 import { types } from 'node:util';    // → types.isNativeError
 const util = require('node:util');    // → util.types.isNativeError
 ```
-
-This unified approach ensures your codemods work correctly regardless of how developers import Node.js modules in their projects.
```
AugustinMauroy, line 234: ah right !

utils/src/ast-grep/update-binding.ts
```diff
@@ -0,0 +1,351 @@
...
+							kind: "lexical_declaration",
+						},
+					},
+				},
+				{
+					kind: "identifier",
+					inside: {
+						kind: "import_clause",
+					},
+				},
```
AugustinMauroy, line 97: Wouldn't it be simpler to have two ast grep query ?

brunocroh: Hmm, did you say something like it? 

```js
let identifiers = []
const importClauses = node.findAll(import_clause)
for(const importClause of importClauses) {
  const _identifiers = importClause.findAll(identifier)
  if(_identifiers.length) {
    identifiers = identifiers.concat(_identifiers)
  }
}
```

PR #186, opened 2025-08-19
Feat/util print to console log

### LINKED ISSUE #124: feat: handle `util.print()` depreciation
## Description

Since `util.print()` is deprecated (DEP0026) and has reached End-of-Life status, we should provide a codemod to replace it.

- The codemod should replace all instances of `util.print()` with `console.log()`.
- The codemod should update import/require statements to replace `node:util` with `node:console` when `print` is the only util function being used.
- If the `node:util` import is destructured and only contains `print`, it should be replaced with `node:console` destructured import.
- If there are other util functions being used, the codemod should add a separate `node:console` import and keep the existing `node:util` import.

## Additional Information

Note that `util.print()` was removed in Node.js v12.0.0 and replaced with `console.log()`. The functionality is identical, but `console.log()` is the standard way to output text to stdout. `util.print()` was deprecated because it was redundant with the more widely used `console.log()` API.

### Case 1: Destructured import with only print

**Before:**
```js
const { print } = require('node:util');

print('Hello world');
```

**After:**
```js
console.log('Hello world');
```

### Case 2: Namespace import

**Before:**
```js
const util = require('node:util');

util.print('Hello world');
```

**After:**
```js
console.log('Hello world');
```

### Case 3: Mixed usage with other util functions

**After**
```js
const { inspect } = require('node:util');

console.log('Hello world');
const formatted = inspect({ foo: 'bar' });
console.log(formatted);
```

### Case 4: ESM import

**Before:**
```js
import { print } from 'node:util';

print('Hello world');
```

**After:**
```js
console.log('Hello world');
```

### Case 5: ESM namespace import

**Before:**
```js
import * as util from 'node:util';

util.print('Hello world');
```

**After:**
```js
console.log('Hello world');
```

### Case 6: Multiple arguments handling

**Before:**
```js
const { print } = require('node:util');

print('Hello', 'world', 123);
```

**After:**
```js
console.log('Hello', 'world', 123);
```

## Refs

- [DEP0026: `util.print()`](https://nodejs.org/api/deprecations.html#dep0026-utilprint)

brunocroh:
Closes #124

the recipe `util-log-to-console-log`

The recipe util-log-to-console-log was only handling [DEP0059](https://nodejs.org/api/deprecations.html#DEP0059), which replaces `util.log` with `console.log`.

However, we have four other similar deprecations, and I think it makes sense to handle them here as well (maybe we should rename the recipe?).

[DEP0026](https://nodejs.org/api/deprecations.html#DEP0026): replace `util.print` with `console.log`

[DEP0027](https://nodejs.org/api/deprecations.html#DEP0027): replace `util.puts` with `console.log`

[DEP0028](https://nodejs.org/api/deprecations.html#DEP0028): replace `util.debug` with `console.error`

[DEP0029](https://nodejs.org/api/deprecations.html#DEP0029): replace `util.error` with `console.error`

### DIFF
--- a/utils/src/ast-grep/remove-lines.ts
+++ b/utils/src/ast-grep/remove-lines.ts
@@ -3,6 +3,8 @@ import type { Range } from "@codemod.com/jssg-types/main";
 export function removeLines(sourceCode: string, ranges: Range[]) {
 	const lines = sourceCode.split("\n");
 
+	// Remove duplicate ranges to prevent attempting to delete the same range multiple times,
+	// which could cause inconsistent behavior
 	const uniqueRemoves = new Map<string, Range>();
 
 	for (const range of ranges) {
--- a/utils/src/ast-grep/remove-lines.test.ts
+++ b/utils/src/ast-grep/remove-lines.test.ts
@@ -124,4 +124,35 @@ describe("remove-lines", () => {
 
 		assert.strictEqual(expected, output);
 	});
+
+	it("Should remove the correct lines when receiving duplicated ranges to remove", () => {
+		const code = dedent`
+			line 1
+			line 2
+			line 3
+			line 4
+		`;
+
+		const expected = dedent`
+			line 1
+			line 4
+		`;
+
+		const range: Range = {
+			start: {
+				line: 1,
+				column: 0,
+				index: 0,
+			},
+			end: {
+				line: 2,
+				column: 0,
+				index: 0,
+			},
+		};
+
+		const output = removeLines(code, [range, range, range, range, range]);
+
+		assert.strictEqual(expected, output);
+	});
 });


brunocroh:
> the DEP0059 is EoL in 24 rest in 12 it's why I think we need to do 2 codemod

Okay, not problem, let me split it then

utils/src/ast-grep/remove-lines.ts
```diff
@@ -3,6 +3,17 @@ import type { Range } from "@codemod.com/jssg-types/main";
 export function removeLines(sourceCode: string, ranges: Range[]) {
 	const lines = sourceCode.split("\n");
 
+	const uniqueRemoves = new Map<string, Range>();
```
AugustinMauroy, line 6: what why (add comment) ? and add test for this case

avivkeller (approved):

Issue #94, opened 2025-06-27
Brainstorm: args update between major


AugustinMauroy:
Between two major releases, there are flags that change. For example, experimental flags that are no longer necessary.
It would be useful to have this kind of codemod to help people with migration.

What are your thoughts?

AugustinMauroy:
after #93 we will be able to create :
- a utility to modify a command line in scripts (package.json)
- update script in a GHA 
- update script in a Gitlab action

AugustinMauroy:
what is the best approach ?
- using JS api with JS parser (maybe strange side effects with `codemod@next`)
-  using ast-grep api 
- other ???

JakobJingleheimer:
I expect probably ast-grep, but other libs may be better suited to specific tasks. So I think default to ast-grep, and consider others where ast-grep doesn't seem an ideal fit?

AugustinMauroy:
Agreed

PR #194, opened 2025-08-29
feat(`utils`): add package.json

AugustinMauroy:
## Description

Add utility to handle changes for `package.json`

## Related issues

Related to #94

### DIFF
--- a/utils/src/ast-grep/package-json.test.ts
+++ b/utils/src/ast-grep/package-json.test.ts
@@ -1,0 +1,159 @@
+import assert from "node:assert/strict";
+import { describe, it } from "node:test";
+import dedent from "dedent";
+import jsonLang from "@ast-grep/lang-json";
+import { registerDynamicLanguage, parse } from "@ast-grep/napi";
+import type { Edit } from "@ast-grep/napi";
+import { getScriptsNode, getNodeJsUsage, replaceNodeJsArgs } from "./package-json.ts";
+
+registerDynamicLanguage({ json: jsonLang });
+
+describe("package-json utilities", () => {
+	describe("getScriptsNode", () => {
+		it("should get the scripts node", () => {
+			const input = dedent`
+				{
+					"scripts": {
+						"test": "echo \\"Error: no test specified\\" && exit 1"
+					}
+				}
+			`;
+
+			const result = getScriptsNode(parse('json', input));
+
+			assert(result);
+			assert.strictEqual(result.children().length, 3); // curly braces + pair + curly braces
+		});
+
+		it("should return empty array if any scripts is present", () => {
+			const input = dedent`
+				{
+					"name": "example-package",
+					"version": "1.0.0"
+				}
+			`;
+
+			const result = getNodeJsUsage(parse('json', input));
+
+			assert.strictEqual(result.length, 0);
+		});
+
+		it("should throw an error if multiple scripts nodes are found", () => {
+			const input = dedent`
+				{
+					"scripts": {
+						"test": "echo \\"Error: no test specified\\" && exit 1"
+					},
+					"scripts": {
+						"start": "node index.js"
+					}
+				}
+			`;
+
+			assert.throws(() => getScriptsNode(parse('json', input)), {
+				message: /Multiple "scripts" fields found/
+			});
+		});
+	});
+
+	describe("getNodeJsUsage", () => {
+		it("should get Node.js usage in scripts", () => {
+			const input = dedent`
+				{
+					"scripts": {
+						"start": "node script.js",
+						"test": "echo \\"Error: no test specified\\" && exit 1"
+					}
+				}
+			`;
+
+			const result = getNodeJsUsage(parse('json', input));
+
+			assert.strictEqual(result.length, 1);
+			assert.strictEqual(result[0].text(), "node script.js");
+		});
+
+		it("should not catch `node_modules`", () => {
+			const input = dedent`
+				{
+					"scripts": {
+						"start": "node_modules/.bin/some-tool",
+						"test": "node another-script.js"
+					}
+				}
+			`;
+
+			const result = getNodeJsUsage(parse('json', input));
+
+			assert.strictEqual(result.length, 1);
+			assert.strictEqual(result[0].text(), "node another-script.js");
+		});
+
+		it("should return empty array if no Node.js usage is found", () => {
+			const input = dedent`
+				{
+					"scripts": {
+						"start": "npm run build",
+						"test": "echo \\"Error: no test specified\\" && exit 1"
+					}
+				}
+			`;
+
+			const result = getNodeJsUsage(parse('json', input));
+
+			assert.strictEqual(result.length, 0);
+		});
+	});
+
+	describe("replaceNodeJsArgs", () => {
+		it("should replace Node.js arguments in scripts", () => {
+			const input = dedent`
+				{
+					"scripts": {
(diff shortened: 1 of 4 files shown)


utils/src/ast-grep/package-json.ts
```diff
@@ -0,0 +1,78 @@
...
+					}
+				}
+			}
+		});
+
+	if (scriptsNodes.length > 1)
+		throw new Error(`Multiple "scripts" fields found in ${packageJsonRootNode.filename()}`);
+
+	return scriptsNodes[0] ?? null;
+};
```
brunocroh, line 29: so the actual implementation returns

```
"scripts": {
  one: "example",
  two: "example",
}
```

that makes be necessary do `result.children()[2]`  to get the node with scripts itself, as the first is `scripts` json key and the second one is the column.

If we updated to approach below, it will return a list of nodes with the scripts something like that

```js
[
  one: "example", // as SgNode Object
  two: "example", // as SgnNode Object
]
```

This way, I think it will be easier to manipulate, and throwing an error when not found won’t be necessary, because if there are no scripts, it will just return an empty array.

```suggestion
export const getScriptsNode = (packageJsonRootNode: SgRoot) =>
	packageJsonRootNode.root().findAll({
		rule: {
			kind: "pair",
			inside: {
				kind: "object",
				inside: {
					kind: "pair",
					has: {
						field: "key",
						kind: "string",
						has: {
							kind: "string_content",
							regex: "scripts",
						},
					},
				},
			},
		},
	});
```

utils/src/ast-grep/package-json.test.ts
```diff
@@ -0,0 +1,175 @@
...
+
+registerDynamicLanguage({ json: jsonLang });
+
+describe("package-json utilities", () => {
+	describe("getScriptsNode", () => {
+		it("should get the scripts node", () => {
+			const input = dedent`
+				{
+					"scripts": {
+						"test": "echo \\"Error: no test specified\\" && exit 1"
```
brunocroh, line 17: These double backslashes are escaping the backslash itself instead of the quotes, making this JSON invalid. I think the correct form is with just one.
```suggestion
						"test": "echo \"Error: no test specified\" && exit 1"
```

utils/src/ast-grep/package-json.test.ts
```diff
@@ -0,0 +1,175 @@
...
+					}
+				}
+			`;
+
+			const result = getScriptsNode(parse('json', input));
+
+			assert(result);
+			assert.strictEqual(result.children().length, 3); // curly braces + pair + curly braces
```
brunocroh, line 25: If suggestion on the way `getScriptsNode` works is accepted, it need to be updated to
```suggestion
			assert.strictEqual(result.length, 1); // number of scripts
```

utils/src/ast-grep/package-json.ts
brunocroh: I thought a little bit more about that, and maybe the key `scripts` can be received as a paramter and this function can be something like `getObjectFields` and can be used like


```js
getObjectFields(node, 'scripts')
getObjectFields(node, 'dependencies')
getObjectFields(node, 'workspaces')
getObjectFields(node, 'exports')
```

I do not have sure if getObjectFields is the best name, btw

AugustinMauroy: for now, we will just cover args thing

recipes/correct-ts-specifiers/src/fixtures/e2e/test.ts
brunocroh, line 1: Is this changes part of this PR?

recipes/rmdir/src/workflow.ts
brunocroh, line 1: ditto

utils/src/ast-grep/package-json.ts
```diff
@@ -0,0 +1,126 @@
...
+			for (const arg of args) {
+				const oldArg = arg.text();
+				const newValue = argsToValues[oldArg];
+				if (newValue) {
+					edits.push(arg.replace(newValue));
+				}
+			}
+		}
+	}
+};
```
brunocroh, line 91: I think just returning the edits is a better API than passing in a variable to be mutated in this context.

Because function caller can have full control about what to do with edits.
 
```suggestion
export const replaceNodeJsArgs = (packageJsonRootNode: SgRoot, argsToValues: Record<string, string>) => {
  const edits: Edit[] = []
	for (const usage of getNodeJsUsage(packageJsonRootNode)) {
		const text = usage.text();
		const bashAST = astGrep.parse("bash", text).root();
		const command = bashAST.findAll({ rule: { kind: "command" } });
		for (const cmd of command) {
			const args = cmd.findAll({
				rule: {
					kind: "word",
					not: {
						inside: { kind: "command_name" },
					},
				},
			});
			for (const arg of args) {
				const oldArg = arg.text();
				const newValue = argsToValues[oldArg];
				if (newValue) {
					edits.push(arg.replace(newValue));
				}
			}
		}
	}
	
	return edits
};
```

utils/src/ast-grep/package-json.ts
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Remove Node.js arguments in the "scripts" node of a package.json AST.
+ * @param packageJsonRootNode The root node of the package.json AST.
+ * @param argsToRemove An array of arguments to remove.
+ * @param edits An array to collect the edits made.
+ */
+export const removeNodeJsArgs = (
+	packageJsonRootNode: SgRoot,
+	argsToRemove: string[],
+	edits: Edit[]
```
brunocroh, line 103: ditto
```suggestion
```

recipes/correct-ts-specifiers/src/fixtures/e2e/test.ts
AugustinMauroy: 😅

utils/src/ast-grep/package-json.ts
AugustinMauroy: currently api will be use like that:
```ts
import { removeNodeJsArgs } from "@nodejs/codemod-utils/ast-grep/package-json";
import type { SgRoot, Edit } from "@codemod.com/jssg-types/main";

/**
 * vX-to-vY
 */
export default function transform(root: SgRoot): string | null {
	const rootNode = root.root();
	const edits: Edit[] = [];

	replaceNodeJsArgs(root, { "--experimental-feature": "--feature"}, edits);	
	removeNodeJsArgs(root, ["--foo"], edits);
	if(!edits.length) return null;


	return rootNode.commitEdits(edits);
}
```

With your proposal:
```ts
import { removeNodeJsArgs, replaceNodeJsArgs } from "@nodejs/codemod-utils/ast-grep/package-json";
import type { SgRoot, Edit } from "@codemod.com/jssg-types/main";

/**
 * vX-to-vY
 */
export default function transform(root: SgRoot): string | null {
	const rootNode = root.root();
	const edits: Edit[] = [];

	edits.push(
		replaceNodeJsArgs(rootNode, { "--experimental-feature": "--feature"})
	);
	edits.push(
		removeNodeJsArgs(rootNode, ["--foo"])
	);
	if(!edits.length) return null;


	return rootNode.commitEdits(edits);
}
```

So In overal I didn't have any strong opinion but when you are looking at code format/style my aproach is simpler but your is more descriptive to what happened

JakobJingleheimer (approved):
🙌

brunocroh (approved):

Issue #197, opened 2025-08-31
feat: handle DEP0100 by replacing process.assert() with the assert module


brunocroh:
## Description

This codemod should replace `process.assert()` calls with the [assert module](https://nodejs.org/api/assert.html). It's useful to migrate code that uses the deprecated internal assertion API which has been removed.

It should replace `process.assert()` with require('node:assert')() for robust assertions. It should handle both CommonJS and ESM imports. It should preserve assertion logic and error handling.

### Case 1

**Before:**

```js
process.assert(condition, "Assertion failed");
```

**After:**

```js
const assert = require("node:assert");
assert(condition, "Assertion failed");
```

### Case 2

**Before:**

```js
function validateInput(input) {
  process.assert(typeof input === "string", "Input must be string");
  return input.trim();
}
```

**After:**

```js
const assert = require("node:assert");
function validateInput(input) {
  assert(typeof input === "string", "Input must be string");
  return input.trim();
}
```

### Case 3

**Before:**

```js
process.assert(config.port, "Port must be configured");
```

**After:**

```js
import assert from "node:assert";
assert(config.port, "Port must be configured");
```

## REFS

- [DEP0100: `process.assert()`](https://nodejs.org/api/deprecations.html#DEP0100)

matheusmorett2:
I would like to work on this one.

PR #200, opened 2025-09-06
feat(process-assert-to-assert): introduce

matheusmorett2:
Closes https://github.com/nodejs/userland-migrations/issues/197

### Add codemod recipe `process-assert-to-assert` to handle deprecation [DEP0100](https://nodejs.org/api/deprecations.html#DEP0100).

**Before**
```javascript
const { assert: nodeAssert, env } = require("process");
nodeAssert(condition, "Assertion valid");
```

**After**
```javascript
const assert = require("node:assert");
const { env } = require("process");
assert(condition, "Assertion valid");
```

--- 

### Fix: `remove-binding.ts`
handle scenario where have alias in require

### DIFF
--- a/recipes/process-assert-to-assert/package.json
+++ b/recipes/process-assert-to-assert/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "@nodejs/process-assert-to-assert",
+  "version": "1.0.0",
+  "description": "Handle DEP0100 via transforming `process.assert` to `assert`.",
+  "type": "module",
+  "scripts": {
+    "test": "npx codemod jssg test -l typescript --ignore-whitespace ./src/workflow.ts ./",
+    "test:u": "npx codemod jssg test -l typescript -u ./src/workflow.ts ./"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/nodejs/userland-migrations.git",
+    "directory": "recipes/process-assert-to-assert",
+    "bugs": "https://github.com/nodejs/userland-migrations/issues"
+  },
+  "author": "matheusmorett2",
+  "license": "MIT",
+  "homepage": "https://github.com/nodejs/userland-migrations/blob/main/recipes/process-assert-to-assert/README.md",
+  "devDependencies": {
+    "@codemod.com/jssg-types": "^1.0.3"
+  },
+  "dependencies": {
+    "@nodejs/codemod-utils": "*"
+  }
+}
--- a/recipes/process-assert-to-assert/src/workflow.ts
+++ b/recipes/process-assert-to-assert/src/workflow.ts
@@ -1,0 +1,122 @@
+import type { Edit, Range, Rule, SgNode, SgRoot } from "@codemod.com/jssg-types/main";
+import { getNodeRequireCalls } from "@nodejs/codemod-utils/ast-grep/require-call";
+import { getNodeImportStatements } from "@nodejs/codemod-utils/ast-grep/import-statement";
+import { resolveBindingPath } from "@nodejs/codemod-utils/ast-grep/resolve-binding-path";
+import { removeBinding } from "@nodejs/codemod-utils/ast-grep/remove-binding";
+import { removeLines } from "@nodejs/codemod-utils/ast-grep/remove-lines";
+
+/**
+ * Transforms deprecated `process.assert` usage to the standard `assert` module.
+ * 
+ * Transformations:
+ * 1. Replaces all `process.assert` references with `assert`
+ * 2. Adds the necessary import/require statement if not already present:
+ *    - For ESM or files without require calls: adds `import assert from "node:assert"`
+ *    - For CommonJS (.cjs files or files using require): adds `const assert = require("node:assert")`
+ * 
+ * Examples:
+ * 
+ * Before:
+ * ```js
+ * process.assert(value);
+ * process.assert.strictEqual(a, b);
+ * ```
+ * 
+ * After:
+ * ```js
+ * import assert from "node:assert";
+ * assert(value);
+ * assert.strictEqual(a, b);
+ * ```
+ */
+export default function transform(root: SgRoot): string | null {
+	const rootNode = root.root();
+	const edits: Edit[] = [];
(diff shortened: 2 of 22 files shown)


recipes/process-assert-to-assert/src/workflow.ts
```diff
@@ -0,0 +1,121 @@
...
+					field: 'function',
+					regex: 'require'
+				}
+			}
+		})
+
+		const isCommonJs = root.filename().includes('.cjs')
+
+		if (Boolean(usingRequire) || isCommonJs) {
+			return `const assert = require("node:assert");\n${sourceCode}`
```
AugustinMauroy, line 114: #Use `node:os.EOL` to ensure new lines work correctly for all os

AugustinMauroy (commented):
I didn't see any test for that kind cases

```js
import process from "node:process"
```

But good first pr

recipes/process-assert-to-assert/src/workflow.ts
```diff
@@ -0,0 +1,121 @@
...
+ * process.assert(value);
+ * process.assert.strictEqual(a, b);
+ * ```
+ * 
+ * After:
+ * ```js
+ * import assert from "node:assert";
+ * assert(value);
+ * assert.strictEqual(a, b);
+ * ```
```
brunocroh, line 30: Remove process.assert.strictEqual reference, I think it was added by mistake. This scenario does not exist. Add bold formatting to the before/after text.
```suggestion
 * **Before**:
 * ```js
 * process.assert(value);
 * ```
 * 
 * **After**:
 * ```js
 * import assert from "node:assert";
 * assert(value);
 * ```
```

brunocroh (commented):
Awesome first contribution 🎉, Thank you @matheusmorett2 !

Just small fixes needed

recipes/process-assert-to-assert/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@nodejs/process-assert-to-assert",
+  "version": "1.0.0",
+  "description": "Handle DEP0100 via transforming `process.assert` to `assert`.",
+  "type": "module",
+  "scripts": {
+    "test": "npx codemod jssg test -l typescript --ignore-whitespace ./src/workflow.ts ./",
+    "test:u": "npx codemod jssg test -l typescript -u ./src/workflow.ts ./"
```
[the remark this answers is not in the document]
matheusmorett2: makes sense, thanks

recipes/process-assert-to-assert/src/workflow.ts
brunocroh: My point is that the TypeScript docs say that the code in the before block, `process.assert.strictEqual(a, b)`, will be updated to use `assert.strictEqual`.

However, `process.assert.strictEqual` does not exist, at least I couldn’t find any reference to it.

brunocroh (commented):
Please fix the conflict in the remove-binding file

recipes/process-assert-to-assert/src/workflow.ts
```diff
@@ -0,0 +1,183 @@
...
+
+			const processUsages = rootNode.findAll({
+				rule: {
+					kind: 'member_expression',
+					has: {
+						kind: 'identifier',
+						regex: '^process$'
+					}
+				}
+			});
```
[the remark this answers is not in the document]
brunocroh: I think so

recipes/process-assert-to-assert/src/workflow.ts
```diff
@@ -0,0 +1,207 @@
...
+
+	const requireCalls = getNodeRequireCalls(root, 'process');
+	const importStatements = getNodeImportStatements(root, 'process');
+	const allImports = [...requireCalls, ...importStatements];
+	const processUsages = rootNode.findAll({
+		rule: {
+			kind: 'member_expression',
+			has: {
+				kind: 'identifier',
+				regex: '^process$',
```
brunocroh: nit: this can be simplified to
```suggestion
				pattern: 'process',
```

recipes/process-assert-to-node-assert/src/workflow.ts
```diff
@@ -0,0 +1,230 @@
...
+
+	sourceCode = removeLines(sourceCode, linesToRemove);
+
+	if (edits.length === 0 && linesToRemove) return sourceCode;
+
+	const alreadyRequiringAssert = getNodeRequireCalls(root, 'assert');
+	const alreadyImportingAssert = getNodeImportStatements(root, 'assert');
+
+	if (alreadyRequiringAssert.length || alreadyImportingAssert.length)
+		return sourceCode;
```
JakobJingleheimer:
```suggestion
	if (alreadyRequiringAssert.length || alreadyImportingAssert.length) return sourceCode;
```

recipes/process-assert-to-node-assert/src/workflow.ts
```diff
@@ -0,0 +1,230 @@
...
+	// `createRequire`).
+	if (usingImport || isMjsFile) {
+		return `import assert from "node:assert";${EOL}${sourceCode}`;
+	}
+
+	if (usingRequire || isCjsFile) {
+		return `const assert = require("node:assert");${EOL}${sourceCode}`;
+	}
+
+	const packageJsonPath = join(process.cwd(), 'package.json');
```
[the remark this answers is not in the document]
JakobJingleheimer: I think since this check is last it should be okay: the migration has already determined that the file contents is ambiguous; this will forcibly disambiguate the file, but that's probably okay.

I think best to add a note around here though that the sequence of checks needs to stay:

1. pre-existing `import` statement(s)
2. pre-existing `require()` calls
3. authoritative file extension
4. pjson.type

JakobJingleheimer (commented):
Thanks for this! It's looking pretty good :)

recipes/process-assert-to-node-assert/src/workflow.ts
AugustinMauroy: the auto format of vscode undo this change

PR #211, opened 2025-09-18
CI: Ensure all new code follows Biome formatting rules.

brunocroh:
Add the following commands:

`format` → command was added to the CI pipeline that runs on PR creation
`format:fix` → was added to the pre-commit scripts.

The only files that I manually change something was `.github/workflows/ci.yml` and `package.json`, all other changes were result from `npx biome format --fix`


brunocroh:
Let await for tooling from [oxc](https://github.com/oxc-project/oxc)

PR #218, opened 2025-09-28
fix: handle property accesses cases on resolve-binding-path

brunocroh:
When a module is imported with property access after require

```js
const bind = require("module").a.b.c
const result = resolveBindingPath(node, "$.a.b.c")
```

It resolve as `bind.a.b.c` However, since the property is accessed immediately after **require**, it should instead resolve as `bind`, because the value of `c` is directly assigned to the bind constant.

This changes fixes that behavior.

### DIFF
--- a/utils/src/ast-grep/resolve-binding-path.ts
+++ b/utils/src/ast-grep/resolve-binding-path.ts
@@ -68,6 +68,31 @@ function resolveBindingPathRequire(node: SgNode<Js>, path: string) {
 		});
 	}
 
+	const propertyAccesses = activeNode.findAll({
+		rule: {
+			kind: "property_identifier",
+			inside: {
+				kind: "member_expression",
+			},
+		},
+	});
+
+	if (propertyAccesses.length) {
+		const pathArr = path.split(".");
+		let newPath = ["$"];
+		let i = 0;
+
+		for (; i < propertyAccesses.length; i++) {
+			// pathArr[i+1] to skip the first element (which is $) that was used for binding replacement
+			if (propertyAccesses[i]?.text() !== pathArr[i + 1]) {
+				return undefined;
+			}
+		}
+
+		// Get the remaining path that was not used in propertyAccesses
+		path = newPath.concat(pathArr.splice(i + 1)).join(".");
+	}
+
 	activeNode = activeNode.child(0);
 
 	if (activeNode?.kind() === "identifier") {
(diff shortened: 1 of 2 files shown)


AugustinMauroy (commented):
is it possible to have handling  something like :
```js
import foo from 'node:foo'

foo.bar.functionToModify();
```

AugustinMauroy:
> > is it possible to have handling something like :
> > ```js
> > import foo from 'node:foo'
> > 
> > foo.bar.functionToModify();
> > ```
> 
> I think this test already covers this scenario, doesn’t it?
> 
> https://github.com/nodejs/userland-migrations/blob/485cdc18c3ad9706dc55226585bd5e6299d604ed/utils/src/ast-grep/resolve-binding-path.test.ts#L26-L41

Yeah but for esm

utils/src/ast-grep/resolve-binding-path.ts
```diff
@@ -68,6 +68,31 @@ function resolveBindingPathRequire(node: SgNode<Js>, path: string) {
...
+			kind: "property_identifier",
+			inside: {
+				kind: "member_expression",
+			},
+		},
+	});
+
+	if (propertyAccesses.length) {
+		const pathArr = path.split(".");
+		let newPath = ["$"];
```
JakobJingleheimer, line 82: nit: I think `newPath` is actually constant?

utils/src/ast-grep/resolve-binding-path.ts
```diff
@@ -68,6 +68,31 @@ function resolveBindingPathRequire(node: SgNode<Js>, path: string) {
...
+		let i = 0;
+
+		for (; i < propertyAccesses.length; i++) {
```
JakobJingleheimer, line 85: `i` will actually be scoped to the parent of `for` automatically, so extracting it like this is unnecessary, but perhaps you did it to make it more obvious that the parent scoping is necessary/intended?

JakobJingleheimer (approved):

utils/src/ast-grep/resolve-binding-path.ts
brunocroh: Good, let me update it

utils/src/ast-grep/resolve-binding-path.ts
brunocroh: I’m not sure if I understood it correctly, but I think that if I declare it inside the for header, it’ll only be available within the for loop’s scope, as the TypeScript compiler warns in the image below.

Screenshot 2025-10-08 at 15 59 38

PR #231, opened 2025-10-09
feat(DEP0187): Passing invalid argument types to fs.existsSync

techflare641:
This PR adds a new codemod to handle DEP0187 deprecation warnings by validating and converting invalid argument types to `fs.existsSync()`.

### Changes
- Implements automatic conversion of invalid argument types (numbers, objects, null, etc.) to valid types (string, Buffer, or URL)
- Handles both CommonJS (`require`) and ESM (`import`) syntax
- Supports namespace imports, destructured imports, and direct usage
- Adds type checks for variable arguments
- Includes comprehensive test cases covering various scenarios

### Examples
- Literal values: `fs.existsSync(123)` → `fs.existsSync(String(123))`
- Null values: `fs.existsSync(null)` → `fs.existsSync(String(null || ''))`
- Variables: Adds runtime type check before the call
- Valid arguments (strings, Buffer, URL) are left unchanged

### DIFF
--- a/recipes/fs-existssync-valid-args/src/workflow.ts
+++ b/recipes/fs-existssync-valid-args/src/workflow.ts
@@ -65,7 +65,7 @@ export default function transform(root: SgRoot<JS>): string | null {
 				if (edit) {
 					edits.push(edit);
 				}
-			} else if (isLiteralOrExpression(argKind)) {
+			} else if (LITERAL_OR_EXPRESSION_KINDS.includes(argKind)) {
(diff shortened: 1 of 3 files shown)


JakobJingleheimer:
Based on the examples in the OP, I think these are not something we should do: node itself could do this but doesn't, presumably, because there is too much assumption, and in some cases, those are smells of real problems (this would sweep those real problems under the rug, making the situation worse / harder to find).

If I've misunderstood, please do let me know.

PR #247, opened 2025-10-21
chore(VSC): add codemod MCP

AugustinMauroy:
https://docs.codemod.com/model-context-protocol
https://code.visualstudio.com/docs/copilot/customization/mcp-servers

As long as people are using GH copilot, we might as well have quality context.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,5 +12,8 @@
   "[markdown]": {
     "editor.wordWrap": "off"
   },
+  "[typescript][javascript]": {
+    "editor.defaultFormatter": "biomejs.biome"
+  },
   "typescript.experimental.useTsgo": true
 }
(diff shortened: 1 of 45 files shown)


brunocroh (commented):
I’m not sure if adding editor settings to the repo is a good idea, since everyone might use a different setup.

Maybe just linking to the documentation on how to set it up would be enough?

AugustinMauroy:
> I’m not sure if adding editor settings to the repo is a good idea, since everyone might use a different setup.
> 
> Maybe just linking to the documentation on how to set it up would be enough?

If I remember correctly, stackOverflow conducted a survey that placed VSCode at 70% of the market. So adding theses settings isn't crazy in my opinion.

brunocroh (approved):

Issue #257, opened 2025-10-31
util-log-to-console-log has an incorrect description


mureinik:
The codemod's description states that it (bolding added by me for emphasis)

> This recipe transforms the usage of **log.util()** to use console.log().

`log.util` is a mistake, and the description should state that

> This recipe transforms the usage of **util.log()** to use console.log().

PR #258, opened 2025-10-31
Fix util-log-to-console-log README

mureinik:
This recipe fixes usage of `util.log`, not `log.util` (which isn't even a thing in core Node.js).

Fixes: #257


AugustinMauroy:
Thanks @mureinik for your first contribution 🎉

JakobJingleheimer:
Thanks!

PR #256, opened 2025-10-30
feat(chalk-to-util-styletext): introduce

richiemccoll:
## Related issue

Fixes https://github.com/nodejs/userland-migrations/issues/198

## Test cases

- [x] basic  
- [x] advanced modifiers
- [x] bg styles
- [x] chained styles
- [x] complex chaining
- [x] commonjs require
- [x] ESM dynamic import
- [x] mixed imports and usage
- [x] warn when un-migratable chalk api is found
- [x] different importing methods
  - [x] `const mychalk = require('chalk');`
  - [x] `const { red } = require('chalk');` _maybe hard_
  - [x] `const { red; foo }. = require("chalk");`
  - [x] `const foo = await import('chalk');`
  - [x] `const { red } = await import('chalk')`
  - [x] `import * as chalk from 'chalk'`
  - [x] `import { red } from 'chalk';`
  - [x] `import { yellow as y } from 'chalk';`
- [x] dependency removal(?)

PR #270, opened 2025-11-08
feat(`dep-manager`): introduce 

AugustinMauroy:


AugustinMauroy:
Why not good idea never heard of theses package. But I need to see if there are compatible with JSSG

AugustinMauroy:
close in favor of #256

PR #274, opened 2025-11-13
clean+improve

AugustinMauroy:
- remove remaining `@next`
- improve ast query and reduce regex complexity 
- remove custom way to catch dynamic import + force right indentation based on editor config
- remove remaning `tsconfig.json`

### DIFF
--- a/recipes/correct-ts-specifiers/src/workflow.ts
+++ b/recipes/correct-ts-specifiers/src/workflow.ts
@@ -1,14 +1,14 @@
 import module from 'node:module';
+import type { RegisterHooksOptions } from 'node:module';
 
 import { type Api, api } from '@codemod.com/workflow';
 import type { Helpers } from '@codemod.com/workflow/dist/jsFam.d.ts';
 
 import { mapImports } from './map-imports.ts';
 import type { FSAbsolutePath } from './index.d.ts';
-
 import * as aliasLoader from '@nodejs-loaders/alias/alias.loader.mjs';
 
-module.registerHooks(aliasLoader);
+module.registerHooks(aliasLoader as RegisterHooksOptions);
 
 export async function workflow({ contexts, files }: Api) {
 	await files(globPattern).jsFam(processModule);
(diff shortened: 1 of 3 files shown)


recipes/node-url-to-whatwg-url/src/url-format.ts
```diff
@@ -103,44 +109,45 @@ const getValue = (pair: SgNode<JS>): V | undefined => {
...
+										{ kind: 'number' },
+										{ kind: 'true' },
+										{ kind: 'false' },
 									],
 								},
 							}),
 						);
-						if (qkeyNode && qvalLiteral !== undefined) list.push([qkeyNode.text(), qvalLiteral]);
+						if (qkeyNode && qvalLiteral !== undefined)
+							list.push([qkeyNode.text(), qvalLiteral]);
```
JakobJingleheimer: I think this is not better: either keep inline or it needs `{}`

recipes/node-url-to-whatwg-url/src/url-parse.ts
```diff
@@ -1,37 +1,49 @@
...
+	{
+		key: 'hostname',
+		replaceFn: (
+			base: string,
+			hadSemi: boolean,
+			declKind: 'const' | 'let' | 'var',
+		) => {
+			const kind = declKind === 'var' ? 'let' : declKind;
+
+			return `${kind} hostname = ${base}.hostname.replace(/^\\[|\\]$/, '')${hadSemi ? ';' : ''}`;
```
JakobJingleheimer: nit: this regex can be extracted/moved above with the others to save re-constructing it every invocation.

JakobJingleheimer (approved):
Thanks for the hygiene sweep! A couple nits, and strangely the type error—possibly due to removing `recipes/correct-ts-specifiers/src/nodejs.d.ts`? (if that's not it, I can try to investigate it next week)

recipes/node-url-to-whatwg-url/src/url-format.ts
AugustinMauroy: + 1 for `(}` since inline is +80 in collum

recipes/correct-ts-specifiers/src/workflow.ts
```diff
@@ -1,14 +1,14 @@
...
 import { type Api, api } from '@codemod.com/workflow';
 import type { Helpers } from '@codemod.com/workflow/dist/jsFam.d.ts';
 
 import { mapImports } from './map-imports.ts';
 import type { FSAbsolutePath } from './index.d.ts';
-
 import * as aliasLoader from '@nodejs-loaders/alias/alias.loader.mjs';
 
-module.registerHooks(aliasLoader);
+module.registerHooks(aliasLoader as RegisterHooksOptions);
```
JakobJingleheimer: very nit: does this work with `satisfies` instead of `as`? if so, that's preferable

```suggestion
module.registerHooks(aliasLoader satisfies RegisterHooksOptions);
```

JakobJingleheimer (approved):

recipes/correct-ts-specifiers/src/workflow.ts
JakobJingleheimer: Ah. Go ahead with the `as` and I'll try to look into the type issue tomorrow 🙂

AugustinMauroy: Okay !

PR #278, opened 2025-11-19
feat(DEP0121): `net._setSimultaneousAccepts()`

### LINKED ISSUE #173: feat: handle `net._setSimultaneousAccepts()` deprecation
## Description

Since `net._setSimultaneousAccepts()` is deprecated (DEP0121) and has reached End-of-Life status, we should provide a codemod to replace it.

- The codemod should remove calls to `net._setSimultaneousAccepts()` as it was an internal API.
- The codemod should handle both CommonJS and ESM imports.
- The codemod should suggest alternative approaches for connection handling if needed.
- The codemod should remove this internal API usage completely.

## Additional Information

Note that `net._setSimultaneousAccepts()` was removed in Node.js v24.0.0. The function was deprecated because it was an internal API that was never intended for public use. This function controlled how many connections could be accepted simultaneously on Windows, but this behavior is now handled automatically by Node.js.

The `net._setSimultaneousAccepts()` function was part of the internal networking implementation and its usage was not recommended or documented for public consumption.

### Example 1: Remove internal API call

**Before:**
```js
const net = require("node:net");

net._setSimultaneousAccepts(false);
const server = net.createServer();
```

**After:**
```js
const net = require("node:net");

const server = net.createServer();
```

### Example 2: Remove from server initialization

**Before:**
```js
const net = require("node:net");

function createServer() {
  net._setSimultaneousAccepts(true);
  return net.createServer((socket) => {
    // handle connection
  });
}
```

**After:**
```js
const net = require("node:net");

function createServer() {
  return net.createServer((socket) => {
    // handle connection
  });
}
```

### Example 3: Remove from module setup

**Before:**
```js
const net = require("node:net");

net._setSimultaneousAccepts(false);
module.exports = {
  createServer: () => net.createServer()
};
```

**After:**
```js
const net = require("node:net");

module.exports = {
  createServer: () => net.createServer()
};
```

### Example 4: Remove from application startup

**Before:**
```js
const net = require("node:net");

function initializeApp() {
  net._setSimultaneousAccepts(true);
  const server = net.createServer();
  server.listen(3000);
}
```

**After:**
```js
const net = require("node:net");

function initializeApp() {
  const server = net.createServer();
  server.listen(3000);
}
```

### Example 5: ESM import cleanup

**Before:**
```js
import net from "node:net";

net._setSimultaneousAccepts(false);
const server = net.createServer();
```

**After:**
```js
import net from "node:net";

const server = net.createServer();
```

### Example 6: Remove from configuration

**Before:**
```js
const net = require("node:net");

const config = {
  simultaneousAccepts: false,
  port: 8080
};

function setupServer(config) {
  net._setSimultaneousAccepts(config.simultaneousAccepts);
  return net.createServer().listen(config.port);
}
```

**After:**
```js
const net = require("node:net");

const config = {
  port: 8080
};

function setupServer(config) {
  return net.createServer().listen(config.port);
}
```

## Refs

- [DEP0121: `net._setSimultaneousAccepts()`](https://nodejs.org/api/deprecations.html#DEP0121)

vespa7:
Closes #173


AugustinMauroy (commented):
Missing test case for dynamic import

JakobJingleheimer:
> Yes, it’s in draft. It doesn’t need to be corrected yet. I’m waiting for a question to be resolved.

I don't see a question. Is it something we can help with? (Could you point us to it?)

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
```diff
@@ -0,0 +1,309 @@
...
+ *
+ * Note: This was an internal API (DEP0121) that reached End-of-Life in Node.js v24.0.0
+ */
+export default function transform(root: SgRoot<Js>): string | null {
+	const rootNode = root.root();
+	const edits: Edit[] = [];
+	const linesToRemove: Range[] = [];
+
+	const netImportStatements = getAllNetImportStatements(root);
+	if (netImportStatements.length === 0) return null;
```
AugustinMauroy:
```suggestion

	// If no import found we don't process the file
	if (!netImportStatements.length) return null;
```

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
```diff
@@ -0,0 +1,309 @@
...
+				if (key?.text() === propertyName) {
+					const rangeWithComma = expandRangeToIncludeTrailingComma(
+						pair.range(),
+						rootNode.text()
+					);
+					linesToRemove.push(rangeWithComma);
+				}
+			}
+		}
+	}
```
AugustinMauroy: can we have a loop that push to an array then iterate on this array to avoid 3th level of nested loop

AugustinMauroy (commented):
not too bad, good job but there are serval part of the implementation that need to be discussed

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
```diff
@@ -0,0 +1,309 @@
...
+	argNode: SgNode<Js>,
+	linesToRemove: Range[]
+): void {
+	const argKind = argNode.kind();
+
+	if (argKind === 'member_expression') {
+		handleMemberExpressionArgument(rootNode, argNode, linesToRemove);
+	} else if (argKind === 'identifier') {
+		handleIdentifierArgument(rootNode, argNode, linesToRemove);
+	}
```
JakobJingleheimer: nit: switch makes it more clear that its inspecting the same condition for each.

```suggestion
  switch(argKind) {
    case 'member_expression': handleMemberExpressionArgument(rootNode, argNode, linesToRemove); break;
    case 'identifier': handleIdentifierArgument(rootNode, argNode, linesToRemove); break;
	}
```

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
```diff
@@ -0,0 +1,309 @@
...
+function removePropertyFromObjectDeclaration(
+	rootNode: SgNode<Js>,
+	objectName: string,
+	propertyName: string,
+	linesToRemove: Range[]
+): void {
+	const objDeclarations = rootNode.findAll({
+		rule: {
+			any: [
+				{ pattern: `const ${objectName} = $_` },
```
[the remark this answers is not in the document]
JakobJingleheimer: Mmm, I think there's probably a built-in way from ast-grep to get this without needing such a wide pattern.

JakobJingleheimer (commented):
Thanks for this! It's a good first :)

I think this could use some additional test-cases that include more things happening from `node:net` to ensure they don't get broken.

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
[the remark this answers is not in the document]
vespa7: I was thinking in something like this:
const pair = root
	.root()
	.findAll({ rule: { kind: "variable_declarator" } })
	.filter((decl) => decl.field("name")?.text() === objectName)
	.flatMap((decl) => decl.findAll({ rule: { kind: "pair" } }))
	.filter((p) => p.field("key")?.text() === keyName)
	.map((pair) => pair.field("value")?.text())[0]

@brunocroh @JakobJingleheimer @AugustinMauroy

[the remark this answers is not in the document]
Ceres6: I think you're misunderstanding what @vespa7's comment is saying due to the thread, but the equivalent to @brunocroh's code in Alex's code is 
```
const pair = root
	.root()
	.findAll({ rule: { kind: "variable_declarator" } })
	.filter((decl) => decl.field("name")?.text() === objectName)
```
If Brunos' code works it might be a bit better for that chunk, but the complete piece is substituting 
```
	const objDeclarations = rootNode.findAll({
		rule: {
			any: [
				{ pattern: `const ${objectName} = $_` },
				{ pattern: `let ${objectName} = $_` },
				{ pattern: `var ${objectName} = $_` },
			],
		},
	});

	for (const objDecl of objDeclarations) {
		const objectLiterals = objDecl.findAll({ rule: { kind: 'object' } });

		for (const obj of objectLiterals) {
			const pairs = obj.findAll({ rule: { kind: 'pair' } });

			for (const pair of pairs) {
				const key = pair.child(0);
				if (key?.text() === propertyName) {
					const rangeWithComma = expandRangeToIncludeTrailingComma(
						pair.range(),
						rootNode.text()
					);
					linesToRemove.push(rangeWithComma);
				}
			}
		}
	}
``` 
And if you got that correctly I'm not sure how the new piece looks more verbose. Following on Bruno's proposal for the whole snippet, maybe something like
```
 const pairs = root
        .root()
        .findAll({
	        rule: {
		        kind: "variable_declarator",
		        has: {
			        kind: "identifier",
			        field: "name",
			        pattern: objectName,
		        },
	        },
        })
        .flatMap((decl) =>
	        decl.findAll({
		        rule: {
			        kind: "pair",
			        has: {
				        field: "key",
				        regex: keyName,
			        },
		        },
	        }),
        );
```
You can break it in the middle to give explanatory names to the steps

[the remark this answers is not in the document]
JakobJingleheimer: Ohhhhh, sorry. I didn't realise these were different pieces of a larger whole. In that case, yes that's fine :)

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
```diff
@@ -0,0 +1,305 @@
...
+	const linesToRemove: Range[] = [];
+
+	const netImportStatements = getAllNetImportStatements(root);
+
+	// If no import found we don't process the file
+	if (!netImportStatements.length) return null;
+
+	for (const statement of netImportStatements) {
+		processNetImportStatement(rootNode, statement, linesToRemove, edits);
+	}
```
JakobJingleheimer: nit: The length check is unnecessary: `getAllNetImportStatements` (and `@nodejs/codemod-utils:getModuleDependencies`) always return an array, so the `for…of` won't break and will simply do nothing when the array is empty; then the check right after will handle aborting, affectively achieving the same thing with less code.

```suggestion
	for (const statement of getAllNetImportStatements(root)) {
		processNetImportStatement(rootNode, statement, linesToRemove, edits);
	}
```
```suggestion
	for (const statement of getModuleDependencies(root)) {
		processNetImportStatement(rootNode, statement, linesToRemove, edits);
	}
```

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
```diff
@@ -0,0 +1,305 @@
...
+	const endPos = range.end.index;
+
+	if (endPos < sourceText.length && sourceText[endPos] === ',') {
+		return {
+			start: range.start,
+			end: {
+				...range.end,
+				index: endPos + 1,
+				column: range.end.column + 1
+			}
```
JakobJingleheimer: nit (formatting/linting)

```suggestion
				column: range.end.column + 1,
				index: endPos + 1,
			},
```

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
```diff
@@ -0,0 +1,305 @@
...
+
+		if (parent.kind() === 'program') {
+			return current;
+		}
+
+		current = parent;
+	}
+
+	return null;
+}
```
JakobJingleheimer: I think this can be simplified:

```suggestion
function findTopLevelStatement(node: SgNode<Js>): SgNode<Js> | null {
	let current: SgNode<Js> | null = node;

	while (current = current.parent()) {
		if (current?.kind() === 'program') return current;
	}

	return null;
}
```

I checked with a quick test:

```js
function generateNode(kind) {
	let i = 1;
	return {
		parent() {
			if (i++ < 4) return {
				kind() { return kind },
			};
			return null;
		},
	};
}

findTopLevelStatement(generateNode('not-program'); // null
findTopLevelStatement(generateNode('program'); // { kind: 𝑓 }
```

package.json
```diff
@@ -19,17 +19,17 @@
     "automation",
     "codemod",
     "migrations",
-    "node.js"
+    "node.js"
```
JakobJingleheimer:
```suggestion
    "node.js"
```

JakobJingleheimer (approved):
This looks good to me! I see there is some uncertainty (https://github.com/nodejs/userland-migrations/issues/173#issuecomment-3180973277) about the proper handling for this, so let's get someone from @nodejs/net to chime in before landing.

PS Sorry this took me a while to get back to.

AugustinMauroy (approved):
LGMT !

brunocroh (approved):
LGTM

JakobJingleheimer:
@vespa7 CI is angry about package-lock. I'm not sure why.

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
```diff
@@ -0,0 +1,305 @@
...
+	callNode: SgNode<Js>,
+	linesToRemove: Range[],
+	edits: Edit[]
+): void {
+	const expressionStatement = findParentExpressionStatement(callNode);
+
+	if (expressionStatement) {
+		linesToRemove.push(expressionStatement.range());
+	} else {
+		edits.push(callNode.replace(''));
```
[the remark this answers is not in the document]
JakobJingleheimer: I think this is nonsense…

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
```diff
@@ -0,0 +1,305 @@
...
+				},
+			},
+		})
+		.flatMap((decl) =>
+			decl.findAll({
+				rule: {
+					kind: 'pair',
+					has: {
+						field: 'key',
+						regex: propertyName,
```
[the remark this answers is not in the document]
JakobJingleheimer: This sounds like a good suggestion, but I'm not that familiar with JSSG. I see an example in the docs that seems to suggest this would work: https://docs.codemod.com/jssg/reference#basic-example

@brunocroh thoughts?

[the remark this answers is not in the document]
brunocroh: Here is the documentation about it: https://ast-grep.github.io/guide/rule-config/atomic-rule.html#regex

Yeah, normally I try to suggest and use `pattern` instead of `regex`, but I’m experimenting with this scenario and for some reason, it isn’t working as intended. That’s fine, we can keep it this way for this case.

I’ll try to take a deeper dive into it; maybe it’s an issue on the codemod/ast-grep.

recipes/net-setSimultaneousAccepts-migration/src/workflow.ts
```diff
@@ -0,0 +1,305 @@
...
+					},
+				},
+			}),
+		);
```
brunocroh, line 158: I think this one can be simplified using relational rules, like `inside`.

https://ast-grep.github.io/guide/rule-config/relational-rule.html#relational-rules

```ts
	const pairs = rootNode.findAll({
		rule: {
			kind: 'pair',
			has: {
				field: 'key',
				regex: propertyName,
			},
			inside: {
				kind: 'variable_declarator',
				has: {
					kind: 'identifier',
					field: 'name',
					pattern: objectName,
				},
				stopBy: 'end',
			},
		},
	});

```

Issue #126, opened 2025-07-25
feat: handle `tls.SecurePair` depreciation


AugustinMauroy:
## Description

This codemod should migrate `tls.SecurePair` to `tls.TLSSocket`. It's useful to migrate code that uses the deprecated `tls.SecurePair` class which has been removed.

It should handle both constructor usage and method calls on SecurePair instances.
It should support both `tls.SecurePair` direct usage and destructured imports from `tls` module.

### Case 1

**Before:**
```js
const tls = require('node:tls');

// Using tls.SecurePair constructor
const pair = new tls.SecurePair();
const cleartext = pair.cleartext;
const encrypted = pair.encrypted;

// Direct import
const { SecurePair } = require('node:tls');
const pair2 = new SecurePair();
```

**After:**
```js
const tls = require('node:tls');

// Using tls.TLSSocket instead
const socket = new tls.TLSSocket(socket);
// Note: Direct migration may require additional context-specific changes
// as SecurePair and TLSSocket have different APIs

// Direct import
const { TLSSocket } = require('node:tls');
const socket2 = new TLSSocket(socket);
```

## Case 2

**Before:**
```js
import tls from 'node:tls';

// Using tls.SecurePair constructor
const pair = new tls.SecurePair();
const cleartext = pair.cleartext;
const encrypted = pair.encrypted;

// Direct import
import { SecurePair } from 'node:tls';
const pair2 = new SecurePair();
```

**After:**
```js
import tls from 'node:tls';

// Using tls.TLSSocket instead
const socket = new tls.TLSSocket(socket);
// Note: Direct migration may require additional context-specific changes
// as SecurePair and TLSSocket have different APIs

// Direct import
import { TLSSocket } from 'node:tls';
const socket2 = new TLSSocket(socket);
```

## REFS

- https://nodejs.org/docs/latest/api/deprecations.html#dep0043-tlssecurepair
- https://github.com/nodejs/node/pull/57361
- https://nodejs.org/docs/latest/api/tls.html#class-tlstlssocket

menace31:
@AugustinMauroy  This is my first open-source contribution, and before I get started, I want to make sure I understand the scope correctly:

    1. Should I create a new folder in the recipes directory named tls-securepair-to-tlssocket?
    2. Should the fix also include documentation, or should I focus only on the core functionality?

If you have any recommendations or additional resources (beyond the existing documentation) to help me get started, I’d really appreciate your guidance!

[...]

PR #286, opened 2025-11-27
feat(DEP0043): introduce tls-securepair-to-tlssocket

menace31:
This PR introduces a new codemod to migrate the deprecated tls.SecurePair class to tls.TLSSocket.

Key features of this codemod:

    - Constructor Transformation: Replaces new SecurePair() with new TLSSocket(socket).

    - Smart Renaming: deep-renames variables (e.g., pair → socket, myPair → mySocket) while preserving CamelCase and scope integrity.

    - Import Handling: Supports both CommonJS (require) and ESM (import) patterns, including destructured imports.

    - Cleanup: Removes usage of deprecated properties like .cleartext and .encrypted.

    - Safety: Adds a comment warning users that manual verification is required for the injected socket argument.

Testing: I have added comprehensive test cases covering:

    - CommonJS and ESM import styles.

    - Variable renaming in complex scopes (nested blocks).

    - Destructuring patterns.

Tests were verified locally using a custom test script ensuring input/output matches expectations.


AugustinMauroy:
Pr title should be updated + the architecture need to fit with the good way to create a codemod

menace31:
I've addressed the previous PR issues. The transformation now uses the recommended utilities from @nodejs/codemod-utils for detecting and manipulating imports/requires (getNodeImportStatements, getNodeRequireCalls, updateBinding, removeLines). I've also removed test.sh from Git, updated the documentation to reference npm test, and all eight tests pass. I hope my work will pass the tests this time.

menace31:
I just corrected the codemod error and ran all tests with npm run test. It should work correctly this time. Thank you for your patience

AugustinMauroy (commented):
wow better still need to address the comment about the readme from my previous review batch https://github.com/nodejs/userland-migrations/pull/286/files#r2592245342

instead it's good

AugustinMauroy:
> I could find a Windows laptop and reproduce the error. I modified the workflow.ts file to normalize the output for Windows. This code was made by AI. I hope this is fine. But now tests pass on linux and Windows too.

If you use AI you can use codemod MCP to have a cool result https://docs.codemod.com/model-context-protocol

menace31:
Since the tests passed, I think this is ready to be reviewed.

AugustinMauroy (commented):
Not bad small point that look strange. Maybe add warn when adding the `socket` so used know where to verify

recipes/tls-securepair-to-tlssocket/src/workflow.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (change?.edit) edits.push(change.edit);
+    if (change?.lineToRemove) linesToRemove.push(change.lineToRemove);
+  }
+
+  const newExpressions = rootNode.findAll({
+    rule: {
+      kind: "new_expression",
+      has: {
+        any: [
+          { kind: "member_expression", has: { field: "property", regex: "^SecurePair$" } },
```
brunocroh: nit: The match can be simplified by using a pattern instead of a regex.
```suggestion
          { kind: "member_expression", has: { field: "property", pattern: "SecurePair" } },
```

recipes/tls-securepair-to-tlssocket/src/workflow.ts
```diff
@@ -0,0 +1,114 @@
...
+          { kind: "identifier", regex: "^SecurePair$" }
```
brunocroh: nit: The match can be simplified by using a pattern instead of a regex.
```suggestion
          { kind: "identifier", pattern: "SecurePair" }
```

recipes/tls-securepair-to-tlssocket/src/workflow.ts
```diff
@@ -0,0 +1,114 @@
...
+    const declarator = getClosest(node, ["variable_declarator"]);
+    if (declarator) {
+      const idNode = declarator.field("name");
+      if (idNode) {
+        const oldName = idNode.text();
+        let newName = "socket";
+        if (oldName !== "pair" && oldName !== "SecurePair") {
+          if (oldName.includes("Pair")) newName = oldName.replace("Pair", "Socket");
+          else if (oldName.includes("pair")) newName = oldName.replace("pair", "socket");
+        }
```
brunocroh: This code block feels a little odd to me, could you explain the idea behind it? It seems very dependent on `oldName`, which feels a bit tricky. Why do we need it? Maybe we could do this in another way?

recipes/tls-securepair-to-tlssocket/src/workflow.ts
```diff
@@ -0,0 +1,114 @@
...
+          if (parentKind === "member_expression") {
+            const property = parent.field("property");
+            if (property && property.id() === ref.id()) continue;
+          }
+          if (parentKind === "import_specifier" || parentKind === "shorthand_property_identifier_pattern") continue;
+          if (ref.id() === idNode.id()) continue;
+
+          edits.push(ref.replace(newName));
+        }
+      }
```
brunocroh: Could you please explain what this block is doing and why it’s necessary? I can’t quite understand it just by reading the code.

brunocroh (commented):
Could you please check your Biome configuration? I think the code isn’t following our code style.

recipes/tls-securepair-to-tlssocket/src/workflow.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  const sourceCode = rootNode.commitEdits(edits);
+
+  let output = removeLines(sourceCode, linesToRemove) ?? "";
+
+  return output;
```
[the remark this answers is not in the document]
menace31: I pushed this quickly to move the code from Windows to Linux for testing. It looks a bit strange right now because it doesn't contains the original logic for Windows character normalization."

menace31:
> Could you please check your Biome configuration? I think the code isn’t following our code style.

I ran this command:

 npx @biomejs/biome check --write

answere" Checked 2 files in 37ms. No fixes applied."

I hope it was the right one to check my code. But To improve readability I added some comments to explain my code

recipes/tls-securepair-to-tlssocket/tests/input/basic-test-file-3.js
```diff
@@ -0,0 +1,13 @@
+const tls = require('node:tls');
+const fs = require('fs'); // Code non lié
+
+function createSecureConnection() {
+  // Using tls.SecurePair constructor
+  const pair = new tls.SecurePair();
+  
+  // Ces lignes doivent disparaître
```
brunocroh: nit: remove or translate it to english please?

```suggestion
  // Ces lignes doivent disparaître
```

brunocroh (commented):
Add a test like the example below to cover cases where the import name is renamed.

```ts
import { SecurePair as Test } from 'node:tls'

// Using tls.SecurePair constructor
const pair = new Test();
const cleartext = pair.cleartext;
const encrypted = pair.encrypted;

// Direct import
const { SecurePair } = require('node:tls');
const pair2 = new Test();
```

brunocroh:
This entire implementation assumes that a variable called socket exists and is required for it to work, but I don’t think that’s true in most cases.

```ts
const tls = require('node:tls');

// Using tls.SecurePair constructor
const pair = new tls.SecurePair();
const cleartext = pair.cleartext;
const encrypted = pair.encrypted;

// Direct import
const { SecurePair } = require('node:tls');
const pair2 = new SecurePair();
```

The examples in the repository, as well as the before/after examples in issue #126, are incomplete. We need more details on how to identify the socket stream in the old code in order to update it to use the correct variable.

A real-world example will probably be needed. If someone has one, it would be very helpful. If not, it would be good for us to create one for testing. I don’t have much experience with these APIs, but I can create an example in the next few days.

brunocroh:
> Hi team,
> 
> I’m following up on this PR for review. As the final deadline for my open-source course project is approaching, it’s crucial for me to get these changes validated and merged quickly. I’ve addressed all the feedback from previous reviews and hope this version meets the requirements.
> 
> Thanks a lot for your responsiveness!

Unfortunately, open-source projects are not an ideal or controlled environment for school assignments, so we can’t merge changes just to meet a deadline.

As I mentioned in my comment above, the requirements in the issue are not clear enough at the moment. You’ve done a great job up to this point, and I genuinely believe you can share this feedback with your teacher as evidence of a meaningful and valuable contribution.

However, we can’t merge the PR in its current state (and it will probably take a few more weeks before it can be merged). This is a sensitive project, and releasing it as it is could potentially break other projects. Before moving forward, we need more investigation and effort to ensure the changes are safe and well-defined.

Thanks for your understanding.

recipes/tls-securepair-to-tlssocket/src/workflow.ts
```diff
@@ -0,0 +1,159 @@
...
+	output = output.replace(/\r\n/g, '\n').replace(/\r/g, '\n');
+	output = output.replace(/(^.*\S)[ \t]+$/gm, '$1');
+	output = output.replace(/^\uFEFF/, '');
+
+	const eol =
+		typeof process !== 'undefined' && process.platform === 'win32'
+			? '\r\n'
+			: '\n';
+	output = output.replace(/\n/g, eol);
+	if (output.endsWith(eol)) output = output.slice(0, -eol.length);
```
brunocroh: I don’t think it’s worth keeping this there if it’s only for Windows tests. We can find another solution for that.

menace31:
Sorry for the delay, I was in an exam session and on vacation.
To be honest, I don’t think I’ll be active enough to keep this issue.
If I can make a few changes to leave clean code for the next contribution, I’d be happy to do so. Otherwise, I think we can leave the code as it is.

Thank you for guiding me through this contribution.

AugustinMauroy:
@brunocroh you take over this pr is there are any news ?

brunocroh:
> @brunocroh you take over this pr is there are any news ?

I was doing some tests on this, and from what I understand, this codemod and DEP0064 need to be handled together.

The reason is that, to know whether something is a SecurePair and migrate it to TLSSocket, you need to know that it comes from tls.createSecurePair(). The other codemod is the one that replaces that call.

So what needs to be done is to perform both migrations at the same time. That way, you can replace how SecurePair is created and update its usage in a single migration, which keeps everything consistent.

They can't be run independently.

I created here a fixture, that can be used to confirm everything is right, after run the codemods.
https://github.com/brunocroh/node-tls-secure-pair-deprecation-spike

AugustinMauroy:
ohhh okay !

PR #296, opened 2025-12-10
feat(`tape-to-node-test`): introduce 

### LINKED ISSUE #260: spec: `tape-to-node-test-runner`
## Description

We should provide a codemod to migrate projects from the Tape test framework to the native Node.js test runner ("node:test").

- The codemod should replace Tape’s import/require with node:test and assert/strict equivalents.
- The codemod should translate common Tape assertion APIs to assert/strict equivalents (e.g., equality, deep equality, throws/doesNotThrow, error checks, pass/ok).
- The codemod should preserve test structure, including nested subtests, skip/todo flags, and asynchronous patterns (promise-based and async/await).
- The codemod should convert teardown hooks and timeouts to node:test options and lifecycle hooks where possible.
- The codemod should annotate or safely fall back for behaviors that don’t have a one-to-one mapping (e.g., plan-based assertion counts, callback-style `t.end` semantics).
- The codemod should be designed with an iterative equivalence-testing method (see “Important points”) and an AST transformation that is maintainable and scoped.
- The codemod should include a mapping matrix that clearly documents feature coverage and known limitations.

## Important points

- Iterative comparison method:
  - Maintain a maximal Tape test suite that exercises representative features and edge cases.
  - Maintain an equivalent `node:test` suite that matches behavior and intended outcomes.
  - Use a comparison script that runs both suites in child processes and compares their behavior (e.g., exit codes) to ensure practical parity. Iterate until parity is stable and transformation rules are feasible.
- Scope:
  - Imports and entry points: replace Tape’s entry with `node:test` and `node:assert/strict`.
  - Assertions: map basic and deep equalities, truthiness, throws/doesNotThrow, error checks, array/object deep equality, and pass.
  - Test structure: keep test names, nesting, and subtest hierarchies; preserve skip/todo metadata.
  - Async: preserve async/await and promise-based tests; keep subtest concurrency semantics consistent with node:test where feasible.
  - Lifecycle: convert teardown to node:test lifecycle hooks; translate test-level timeouts to node:test options where possible.
- Expected deliverables:
  - A codemod that applies the outlined transformations across CommonJS and ESM projects.
  - A documented mapping matrix of Tape APIs to node:test/assert equivalents, indicating coverage and caveats.
  - A parity harness that runs both the Tape and node:test suites and reports parity outcomes (e.g., exit code compatibility).
  - A test fixture set demonstrating typical patterns and edge cases used to validate the codemod.
- Steps for contributors (high level):
  - Study the mapping matrix and the two canonical suites to understand intended parity.
  - Implement transformations incrementally, validating after each category (imports, assertions, lifecycle, async, metadata).
  - Extend the suites and parity checks when you discover gaps or edge cases; iterate until stable.
  - Add clear annotations or comments for non-trivial cases where automatic migration is partial or requires follow-up by maintainers.

## Detection Strategy

The codemod should identify Tape usage and migration targets by analyzing import patterns and assertion/test invocations:

- Detect Tape entry points:
  - Default and named imports from the Tape package, as well as CommonJS requires.
  - Local aliases of the Tape test function (e.g., renamed bindings).
- Detect test definitions, nested tests, and metadata:
  - Calls to the Tape test function, including variants with options objects for skip/todo.
  - Nested t.test(...) and its callback parameter binding (e.g., t, st, t1, etc.).
- Detect assertion calls and map categories:
  - Equality: equal/notEqual/strictEqual/notStrictEqual; deepEqual/notDeepEqual; same/notSame.
  - Truthiness: ok/notOk/true/false; pass; error (ifError).
  - Throws: throws/doesNotThrow; message or regex matching.
- Detect lifecycle and timeout:
  - Teardown hooks registered on the test object.
  - Timeout calls; distinguish test-level timeouts from other timers.
- Handle mixed and edge cases:
  - Ensure transformation respects renamed imports/variables.
  - Avoid rewriting unrelated functions named test/assert in other modules.
  - Limit scope to Tape tests in files where Tape is clearly referenced.

### Case 1: Basic equality assertions

**Before:**

```js
import test from "tape";

test("basic equality", (t) => {
  t.plan(4);
  t.equal(1, 1, "equal numbers");
  t.notEqual(1, 2, "not equal numbers");
  t.strictEqual(true, true, "strict equality");
  t.notStrictEqual("1", 1, "not strict equality");
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

test("basic equality", () => {
  assert.equal(1, 1, "equal numbers");
  assert.notEqual(1, 2, "not equal numbers");
  assert.strictEqual(true, true, "strict equality");
  assert.notStrictEqual("1", 1, "not strict equality");
});
```

### Case 2: Deep equality and non-equality

**Before:**

```js
import test from "tape";

test("deep equality", (t) => {
  t.plan(2);
  t.deepEqual({ a: 1 }, { a: 1 }, "objects are deeply equal");
  t.notDeepEqual({ a: 1 }, { a: 2 }, "objects are not deeply equal");
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

test("deep equality", () => {
  assert.deepEqual({ a: 1 }, { a: 1 }, "objects are deeply equal");
  assert.notDeepEqual({ a: 1 }, { a: 2 }, "objects are not deeply equal");
});
```

### Case 3: Truthiness, falsiness, and explicit booleans

**Before:**

```js
import test from "tape";

test("truthiness", (t) => {
  t.plan(4);
  t.ok(true, "true is ok");
  t.notOk(false, "false is not ok");
  t.true(true, "explicitly true");
  t.false(false, "explicitly false");
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

test("truthiness", () => {
  assert.ok(true, "true is ok");
  assert.ok(!false, "false is not ok");
  assert.ok(true, "explicitly true");
  assert.ok(!false, "explicitly false");
});
```

### Case 4: Throws and does-not-throw with message/regex

**Before:**

```js
import test from "tape";

function throwsError() {
  throw new Error("Expected error");
}

test("throws assertions", (t) => {
  t.plan(4);
  t.throws(() => { throw new Error("test error"); }, /test error/, "function throws with message");
  t.doesNotThrow(() => 42, "function does not throw");
  t.throws(throwsError, Error, "throws specific error type");
  t.doesNotThrow(() => ({ success: true }));
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

function throwsError() {
  throw new Error("Expected error");
}

test("throws assertions", () => {
  assert.throws(() => { throw new Error("test error"); }, /test error/, "function throws with message");
  assert.doesNotThrow(() => 42, "function does not throw");
  assert.throws(throwsError, Error, "throws specific error type");
  assert.doesNotThrow(() => ({ success: true }));
});
```

### Case 5: Error presence checks

**Before:**

```js
import test from "tape";

test("error check", (t) => {
  t.plan(1);
  t.error(null, "no error present");
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

test("error check", () => {
  assert.ifError(null);
});
```

### Case 6: Nested tests and subtests

**Before:**

```js
import test from "tape";

test("nested tests", (t) => {
  t.plan(1);
  t.test("inner test 1", (st) => {
    st.plan(1);
    st.equal(1, 1, "inner assertion");
  });
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

test("nested tests", async (t) => {
  await t.test("inner test 1", () => {
    assert.equal(1, 1, "inner assertion");
  });
});
```

### Case 7: Skip and todo metadata

**Before:**

```js
import test from "tape";

test("skip test", { skip: true }, (t) => {
  t.fail("this should be skipped");
});

test("todo test", { todo: true }, (t) => {
  t.fail("this is a todo");
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

test("skip test", { skip: true }, () => {
  assert.fail("this should be skipped");
});

test("todo test", { todo: true }, () => {
  assert.fail("this is a todo");
});
```

### Case 8: Async test using promises or async/await

**Before:**

```js
import test from "tape";

function someAsyncThing() {
  return new Promise((resolve) => setTimeout(() => resolve(true), 50));
}

test("async test with promises", async (t) => {
  t.plan(1);
  const result = await someAsyncThing();
  t.ok(result, "async result is truthy");
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

function someAsyncThing() {
  return new Promise((resolve) => setTimeout(() => resolve(true), 50));
}

test("async test with promises", async () => {
  const result = await someAsyncThing();
  assert.ok(result, "async result is truthy");
});
```

### Case 9: Lifecycle teardown

**Before:**

```js
import test from "tape";

let teardownState = 1;

test("teardown registers and runs after test", (t) => {
  t.plan(1);
  t.teardown(() => { teardownState = 0; });
  t.equal(teardownState, 1, "state before teardown");
});

test("teardown verification in subsequent test", (t) => {
  t.plan(1);
  t.equal(teardownState, 0, "teardown ran and updated state");
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

let teardownState = 1;

test("teardown registers and runs after test", (t) => {
  t.after(() => { teardownState = 0; });
  assert.equal(teardownState, 1, "state before teardown");
});

test("teardown verification in subsequent test", () => {
  assert.equal(teardownState, 0, "teardown ran and updated state");
});
```

### Case 10: Timeouts at test level

**Before:**

```js
import test from "tape";

test("timeout success (no trigger)", (t) => {
  t.plan(1);
  t.timeoutAfter(200);
  setTimeout(() => {
    t.ok(true, "completed before timeout");
  }, 100);
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

test("timeout success (no trigger)", { timeout: 200 }, () => {
  return new Promise((resolve) => {
    setTimeout(() => {
      assert.ok(true, "completed before timeout");
      resolve();
    }, 100);
  });
});
```

### Case 11: Arrays and object structural comparisons

**Before:**

```js
import test from "tape";

test("array assertions", (t) => {
  t.plan(2);
  t.same([1, 2, 3], [1, 2, 3], "arrays are same");
  t.notSame([1, 2, 3], [1, 2, 4], "arrays are not same");
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

test("array assertions", () => {
  assert.deepEqual([1, 2, 3], [1, 2, 3], "arrays are same");
  assert.notDeepEqual([1, 2, 3], [1, 2, 4], "arrays are not same");
});
```

### Case 12: Mixed sync and async assertions in one test

**Before:**

```js
import test from "tape";

function someAsyncThing() {
  return new Promise((resolve) => setTimeout(() => resolve(true), 50));
}

test("mixed sync and async", async (t) => {
  t.ok(true, "sync assertion 1");
  await someAsyncThing();
  t.ok(true, "async assertion");
  t.equal(1 + 1, 2, "sync assertion 2");
});
```

**After:**

```js
import { test } from "node:test";
import assert from "node:assert/strict";

function someAsyncThing() {
  return new Promise((resolve) => setTimeout(() => resolve(true), 50));
}

test("mixed sync and async", async () => {
  assert.ok(true, "sync assertion 1");
  await someAsyncThing();
  assert.ok(true, "async assertion");
  assert.equal(1 + 1, 2, "sync assertion 2");
});
```

## Caveats

- Plan counts: Tape’s explicit `t.plan(n)` has no direct equivalent in node:test; the codemod should annotate such occurrences for manual review or add a clear comment indicating plan semantics were removed.
- Callback-style endings: Tests that rely on `t.end` or callback conclusion patterns require careful handling; the codemod should annotate these for manual follow-up where automatic conversion to async/await or promises isn’t obvious.
- Matcher nuances: Message/regex matching in throws/doesNotThrow is broadly compatible but may have subtle differences; contributors should verify during parity checks.
- Timeout semantics: Converting a runtime-set timeout call to a static test option changes where timeout is expressed; verify that behavior remains equivalent in the suite.
- Teardown scope: Tape teardown is per-test; node:test offers t.after and t.afterEach. The codemod should maintain intended scope and annotate ambiguous cases.
- Output/reporting differences: The comparison harness validates behavior via exit codes and outcomes; textual output formatting (reporters) is not guaranteed to be identical.

## Refs

- Node.js Test Runner (`node:test`): https://nodejs.org/api/test.html
- Node.js Assert (`node:assert/strict`): https://nodejs.org/api/assert.html
- Tape: https://github.com/ljharb/tape

AugustinMauroy:
## Related issue

Close #260


recipes/tape-to-node-test/tests/new-features/expected.js
```diff
@@ -0,0 +1,18 @@
...
+    assert.notEqual(1, '2', 'notLooseEqual');
+    assert.deepEqual({ a: 1 }, { a: '1' }, 'deepLooseEqual');
+    t.diagnostic('this is a comment');
+    assert.ok(!false, 'notOk');
+    // t.end();
+});
+
+// TODO: test.onFinish(() => {
+//     console.log('finished');
+// });
```
ljharb: if this is a TODO for you, then this PR should probably still be a draft; if not, then it would be better to make it fail VERY LOUDLY so that users are aware they have to manually fix something. it's not ok to just silently elide their code :-)

recipes/tape-to-node-test/tests/require-import/expected.js
```diff
@@ -0,0 +1,7 @@
+const { test } = require('node:test');
+const assert = require('node:assert');
+
+test("require test", async (t) => {
+    assert.strictEqual(1, 1);
+    // t.end();
```
ljharb: this line should either be removed, or it should end up with whatever the node test runner has to say "this test is done and if you get more assertions later, it should error". (if it doesn't have that, then the node runner doesn't have sufficient features for a codemod from tape to be appropriate)

ljharb (commented):
(btw the diff would read cleaner if it was "output" instead of "expected", because then "input" would be sorted before "output" instead of the reverse)

some of my comments apply on all the input/expected combos, so i'll post this now before being exhaustive

AugustinMauroy:
> (btw the diff would read cleaner if it was "output" instead of "expected", because then "input" would be sorted before "output" instead of the reverse)
> 
> some of my comments apply on all the input/expected combos, so i'll post this now before being exhaustive

the input/expect structure isn't chose by us it's the design of the codemod tool

recipes/tape-to-node-test/tests/advanced-assertions/expected.js
```diff
@@ -0,0 +1,13 @@
...
+
+test('advanced assertions', async (t) => {
+    assert.throws(() => { throw new Error('fail'); }, /fail/);
+    assert.doesNotThrow(() => { });
+    assert.match('string', /ring/);
+    assert.doesNotMatch('string', /gnirt/);
+    assert.fail('this should fail');
+    assert.ifError(null);
+    assert.ifError(null);
+    // t.end();
```
ljharb:
```suggestion
```

recipes/tape-to-node-test/tests/cjs-destructuring/expected.js
```diff
@@ -0,0 +1,7 @@
+const { test } = require('node:test');
+const assert = require('node:assert');
+
+test('cjs destructuring', async (t) => {
+    assert.ok(true);
+    // t.end();
```
ljharb:
```suggestion
```

recipes/tape-to-node-test/tests/nested-test/expected.js
```diff
@@ -0,0 +1,10 @@
+import { test } from 'node:test';
+import assert from 'node:assert';
+
+test("nested tests", async (t) => {
+	t.plan(1);
+	await t.test("inner test 1", async (st) => {
```
ljharb: the nested test here is sync, there should be no need for any async functions or `await` in this file

recipes/tape-to-node-test/tests/new-features/expected.js
ljharb: that's fine, but can the PR be marked draft in the meantime? a green PR is mergeable, and with this todo i don't think that's accurate.

recipes/tape-to-node-test/tests/plan-and-end/expected.js
```diff
@@ -0,0 +1,11 @@
+import { test } from 'node:test';
+import assert from 'node:assert';
+
+test("basic equality", async (t, done) => {
```
ljharb:
```suggestion
test("basic equality", (t, done) => {
```

recipes/tape-to-node-test/tests/timeout-non-object/expected.js
```diff
@@ -0,0 +1,9 @@
+import { test } from 'node:test';
+import assert from 'node:assert';
+
+const opts = { skip: false };
+
+test('timeout with variable opts', opts, async (t) => {
+	// TODO: Add timeout: `123` to test options manually;
```
ljharb: does the test runner not have an imperative timeout API?

ljharb (commented):
There's a big issue with converting to `assert` - namely, that tape test suites very intentionally do not terminate at a failed assertion, but node's `assert` throws (which prevents running the rest of the assertions).

Does the test runner have a non-throwing assertion API?

recipes/tape-to-node-test/tests/timeout-non-object/expected.js
AugustinMauroy: I didn't found it

AugustinMauroy:
> If there's no equivalent, then probably it doesn't make sense to provide a codemod until there is?

IMO we should warn it on readme and people will be aware and then they know that will change how does it work.
cc @JakobJingleheimer what do you think ?

recipes/tape-to-node-test/src/workflow.ts
```diff
@@ -0,0 +1,578 @@
...
+	notDeepEquals: 'notDeepStrictEqual',
+	notEquivalent: 'notDeepStrictEqual',
+	notDeeply: 'notDeepStrictEqual',
+	isNotDeepEqual: 'notDeepStrictEqual',
+	isNotDeeply: 'notDeepStrictEqual',
+	isNotEquivalent: 'notDeepStrictEqual',
+	isInequivalent: 'notDeepStrictEqual',
+	deepLooseEqual: 'deepEqual',
+	notDeepLooseEqual: 'notDeepEqual',
+};
```
avivkeller:
```suggestion
const ASSERTION_MAPPING = {
	equal: 'strictEqual',
	notEqual: 'notStrictEqual',
	strictEqual: 'strictEqual',
	notStrictEqual: 'notStrictEqual',
	deepEqual: 'deepStrictEqual',
	notDeepEqual: 'notDeepStrictEqual',
	looseEqual: 'equal',
	notLooseEqual: 'notEqual',
	ok: 'ok',
	ifError: 'ifError',
	error: 'ifError',
	throws: 'throws',
	doesNotThrow: 'doesNotThrow',
	match: 'match',
	doesNotMatch: 'doesNotMatch',
	fail: 'fail',
	same: 'deepStrictEqual',
	notSame: 'notDeepStrictEqual',
	// Aliases
	assert: 'ok',
	ifErr: 'ifError',
	iferror: 'ifError',
	equals: 'strictEqual',
	isEqual: 'strictEqual',
	strictEquals: 'strictEqual',
	is: 'strictEqual',
	notEquals: 'notStrictEqual',
	isNotEqual: 'notStrictEqual',
	doesNotEqual: 'notStrictEqual',
	isInequal: 'notStrictEqual',
	notStrictEquals: 'notStrictEqual',
	isNot: 'notStrictEqual',
	not: 'notStrictEqual',
	looseEquals: 'equal',
	notLooseEquals: 'notEqual',
	deepEquals: 'deepStrictEqual',
	isEquivalent: 'deepStrictEqual',
	notDeepEquals: 'notDeepStrictEqual',
	notEquivalent: 'notDeepStrictEqual',
	notDeeply: 'notDeepStrictEqual',
	isNotDeepEqual: 'notDeepStrictEqual',
	isNotDeeply: 'notDeepStrictEqual',
	isNotEquivalent: 'notDeepStrictEqual',
	isInequivalent: 'notDeepStrictEqual',
	deepLooseEqual: 'deepEqual',
	notDeepLooseEqual: 'notDeepEqual',
};
```

The type here can be inferred

recipes/tape-to-node-test/src/workflow.ts
```diff
@@ -0,0 +1,578 @@
...
+		if (ASSERTION_MAPPING[method]) {
+			const newMethod = ASSERTION_MAPPING[method];
+			if (func) {
+				edits.push(func.replace(`assert.${newMethod}`));
+			}
+			continue;
+		}
+
+		switch (method.toLowerCase()) {
+			case 'notok':
```
avivkeller:
```suggestion
			case 'notok':
			case 'false':
```

recipes/tape-to-node-test/src/workflow.ts
```diff
@@ -0,0 +1,578 @@
...
+
+	for (const call of lifecycleCalls) {
+		const { line, column } = call.range().start;
+		const fileName = root.filename();
+		const methodName =
+			call.field('function')?.field('property')?.text() || 'lifecycle method';
+
+		console.warn(
+			`[Codemod] Warning: ${methodName} at ${fileName}:${line}:${column} has no direct equivalent in node:test. Please migrate manually.`,
+		);
```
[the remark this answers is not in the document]
AugustinMauroy: aviv can I invite you to open an issue on test runner repo

brunocroh (approved):

recipes/tape-to-node-test/tests/timeout-non-object/expected.js
JakobJingleheimer: It does via `signal` ([`AbortSignal.timeout()`](https://nodejs.org/api/globals.html#static-method-abortsignaltimeoutdelay)): https://nodejs.org/api/test.html#testname-options-fn

```
test('…', { signal }, () => {
  signal.timeout(123);
});
```

recipes/tape-to-node-test/src/remove-dependencies.ts
JakobJingleheimer: Is this file/util really doing anything?

recipes/tape-to-node-test/src/workflow.ts
```diff
@@ -0,0 +1,578 @@
...
+
+	const testCalls = rootNode.findAll({
+		rule: {
+			kind: 'call_expression',
+			has: {
+				field: 'function',
+				regex: `^${testVarName}(\\.(skip|only))?$`,
+			},
+		},
+	});
```
brunocroh, line 152: this regex will not work in some format scenarios e.g:
```ts
// will work here
test.skip('skipped test', (t) => {
	t.fail('should not run');
});

// not will work because of break row
test
	.skip('skipped test', (t) => {
	t.fail('should not run');
});

// not will work because of optional chaining
test?.skip('skipped test', (t) => {
	t.fail('should not run');
});
```

```suggestion
	const testCalls = rootNode.findAll({
		constraints: {
			METHOD: {
				regex: `^(skip|only)$`,
			},
		},
		rule: {
			kind: 'call_expression',
			any: [
				{
					has: {
						field: 'function',
						pattern: testVarName,
					},
				},
				{
					has: {
						field: 'function',
						all: [
							{
								has: {
									field: 'object',
									pattern: testVarName,
								},
							},
							{
								has: {
									field: 'property',
									pattern: '$METHOD',
								},
							},
						],
					},
				},
			],
		},
	});

```

recipes/tape-to-node-test/tests/test-options/input.js
```diff
@@ -0,0 +1,9 @@
+import test from 'tape';
+
+test.skip('skipped test', (t) => {
+	t.fail('should not run');
+});
+
+test.only('only test', (t) => {
+	t.pass('should run');
+});
```
brunocroh, line 9: add more code format variance
```suggestion
import test from 'tape';

test?.skip('skipped test', (t) => {
	t.fail('should not run');
});

test
	.skip('skipped test', (t) => {
	t.fail('should not run');
});

test.only('only test', (t) => {
	t.pass('should run');
});

test?.only('only test', (t) => {
	t.pass('should run');
});


```

recipes/tape-to-node-test/tests/test-options/expected.js
```diff
@@ -0,0 +1,11 @@
...
+import assert from 'node:assert';
+
+test.skip('skipped test', (t) => {
+	assert.fail('should not run');
+});
+
+test.only('only test', (t) => {
+	// TODO: t.pass('should run')
+// TODO: Manual migration: consider t.diagnostic(message) for informational output, or remove this call.;
+});
```
brunocroh, line 11: add more code format variance

```suggestion
import { test } from 'node:test';
import assert from 'node:assert';

test?.skip('skipped test', (t) => {
	assert.fail('should not run');
});

test
	.skip('skipped test', (t) => {
	assert.fail('should not run');
});

test.only('only test', (t) => {
	// TODO: t.pass('should run')
// TODO: Manual migration: consider t.diagnostic(message) for informational output, or remove this call.;
});

test?.only('only test', (t) => {
	// TODO: t.pass('should run')
// TODO: Manual migration: consider t.diagnostic(message) for informational output, or remove this call.;
});

```

recipes/tape-to-node-test/src/workflow.ts
```diff
@@ -0,0 +1,578 @@
...
+						// 2. test('name', opts, cb) -> update options
+
+						// Filter out punctuation to get actual args
+						const actualArgs = children.filter(
+							(c) =>
+								c.kind() !== '(' &&
+								c.kind() !== ')' &&
+								c.kind() !== ',' &&
+								c.kind() !== 'comment',
+						);
```
brunocroh, line 540:
```suggestion
						const actualArgs = children.filter((c) => c.isNamed() && !c.is('comment'));
```

brunocroh (approved):

avivkeller (approved):
After rebase LGTM

ljharb:
fwiw i still don't think this is a responsible codemod to exist unless it can cover 100% of the features of the thing it's codemodding away from.

mcollina:
I concur with @ljharb 

The reasons why one uses tape are the opposite of why one would use node:test. Both totally valid.

AugustinMauroy:
agree with you. `node:test` ins't a replacement like `tape` is dangerous or slow or iDK.
But this is usefull for people who just wanna to have small unit test for their js function. Since it's help you to remove on dep it's also reduce the impact surface.

let's take an example, if you have a small js package that have typescript, tsx, tape as dep. just with one codemod you can remove one dep. and with `nodejs/correct-ts-specifiers` you maybe able to remove tsx from your dep 

So imo we can land this but improve the docs about "Why and when you should use it "

ljharb:
not a production dep though - a *dev* dep, which means the impact of adding or removing one basically doesn't matter at all.

AugustinMauroy:
that's not right. a supply chain attack can ship a malicious code to your dev dep that poisoned your whole machine. So removing one dev dep reduce risk

PR #304, opened 2025-12-15
doc(readme s): use diff

AugustinMauroy:
basically use diff every where

script to create all tags:

```sh
#!/usr/bin/env bash
set -euo pipefail

# Creates annotated git tags for each recipe package.json under recipes/*/
# Tag format: <package-name>@v<version> (example: @nodejs/fs-access-mode-constants@v1.0.1)

ROOT="$(cd "$(dirname "$0")/.." && pwd)"
cd "$ROOT"

# ensure we are in a git repo
if ! git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
  echo "Not inside a git repository." >&2
  exit 1
fi

# ensure working tree is clean
if ! git diff-index --quiet HEAD --; then
  echo "Working tree is not clean. Commit or stash changes before creating tags." >&2
  git status --porcelain
  exit 1
fi

# Optionally restrict tagging to a curated list of codemods. If empty, all recipes will be used.
CODEMODS=(
  fs-access-mode-constants
  node-url-to-whatwg-url
  util-print-to-console-log
  http-classes-with-new
  create-require-from-path
  tmpdir-to-tmpdir
  buffer-atob-btoa
  types-is-native-error
  repl-builtin-modules
  rmdir
  createCredentials-to-createSecureContext
  crypto-fips-to-getFips
  fs-truncate-fd-deprecation
  slow-buffer-to-buffer-alloc-unsafe-slow
  repl-classes-with-new
  process-main-module
  util-log-to-console-log
  process-assert-to-node-assert
)

files=()
if [ ${#CODEMODS[@]} -gt 0 ]; then
  for r in "${CODEMODS[@]}"; do
    p="recipes/$r/package.json"
    if [ -f "$p" ]; then
      files+=("$p")
    else
      echo "Warning: $p not found; skipping." >&2
    fi
  done
else
  mapfile -t files < <(git ls-files 'recipes/*/package.json')
fi

if [ ${#files[@]} -eq 0 ]; then
  echo "No recipes found (no recipes/*/package.json)." >&2
  exit 1
fi

echo "Found ${#files[@]} package.json files. Will create tags for each package's version."

# Preview list
for f in "${files[@]}"; do
  name=$(node -e "const fs=require('fs');const p=JSON.parse(fs.readFileSync('$f','utf8'));console.log(p.name||'')")
  version=$(node -e "const fs=require('fs');const p=JSON.parse(fs.readFileSync('$f','utf8'));console.log(p.version||'')")
  if [ -z "$name" ] || [ -z "$version" ]; then
    echo "Skipping $f (missing name or version)"
    continue
  fi
  echo "  -> $name@v$version ($f)"
done

read -r -p "Proceed to create annotated tags for these packages? [y/N] " yn
if [[ "$yn" != "y" && "$yn" != "Y" ]]; then
  echo "Aborted by user. No tags created."
  exit 0
fi

# create tags
for f in "${files[@]}"; do
  name=$(node -e "const fs=require('fs');const p=JSON.parse(fs.readFileSync('$f','utf8'));console.log(p.name||'')")
  version=$(node -e "const fs=require('fs');const p=JSON.parse(fs.readFileSync('$f','utf8'));console.log(p.version||'')")
  if [ -z "$name" ] || [ -z "$version" ]; then
    continue
  fi
  tag="$name@v$version"
  # skip if tag already exists
  if git rev-parse -q --verify "refs/tags/$tag" >/dev/null; then
    echo "Tag already exists: $tag — skipping"
    continue
  fi
  msg="chore(release): $tag"
  git tag -a "$tag" -m "$msg"
  echo "Created tag: $tag"
done

echo "All done. To push tags to origin run:"
echo "  git push --tags"
```

### DIFF
--- a/recipes/buffer-atob-btoa/README.md
+++ b/recipes/buffer-atob-btoa/README.md
@@ -8,28 +8,22 @@ Migrates usage of the legacy APIs `buffer.atob()` and `buffer.btoa()` to the cur
 
 ```diff
 - const buffer = require('node:buffer');
++ const data = 'SGVsbG8gV29ybGQh'; // "Hello World!" in base64
 - const data = 'SGVsbG8gV29ybGQh'; // "Hello World!" in base64
 - const decodedData = buffer.atob(data);
-- console.log(decodedData); // Outputs: Hello World!
--
-+ const data = 'SGVsbG8gV29ybGQh'; // "Hello World!" in base64
 + const decodedData = Buffer.from(data, 'base64').toString('binary');
-+ console.log(decodedData); // Outputs: Hello World!
-+
+  console.log(decodedData); // Outputs: Hello World!
 `````
 
 ### Migrating buffer.btoa(data)
 
 ```diff
 - const buffer = require('node:buffer');
++ const data = 'Hello World!';
 - const data = 'Hello World!';
 - const encodedData = buffer.btoa(data);
-- console.log(encodedData); // Outputs: SGVsbG8gV29ybGQh
--
-+ const data = 'Hello World!';
 + const encodedData = Buffer.from(data, 'binary').toString('base64');
-+ console.log(encodedData); // Outputs: SGVsbG8gV29ybGQh
-+
+  console.log(encodedData); // Outputs: SGVsbG8gV29ybGQh
 `````
 
 ## REFS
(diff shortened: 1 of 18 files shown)


JakobJingleheimer (commented):
Thanks for updating these. Buuut a diff is pairs of changes, so the line changed should be immediately followed with the alteration.

AugustinMauroy:
> Thanks for updating these. Buuut a diff is pairs of changes, so the line changed should be immediately followed with the alteration.

@copilot fix theses code snippets

JakobJingleheimer (approved):
🙌

PR #313, opened 2025-12-22
feat(mocha-to-node-test-runner): migrate mocha@8.x to `node:test` (v22, v24+)

### LINKED ISSUE #103: brainstorm: `mocha-to-node-test-runner`
## Description

We need to brainstorm ideas for the `mocha-to-node-test-runner` codemod. It's will support converting Mocha v8 to node v22, v24.

## Important points

- It's must not touch to the imports **except** `mocha` imports and `node:test` imports.
- It's should not convert the CB style from `function()` to arrow functions `() => {}` or vice versa.
- CJS/ESM should both be supported.

### Example 1: Basic Test Case

**Before:**

```js
const assert = require('assert');
describe('Array', function() {
  describe('#indexOf()', function() {
    it('should return -1 when the value is not present', function() {
      const arr = [1, 2, 3];
      assert.strictEqual(arr.indexOf(4), -1);
    });
  });
});
```

**After:**

```js
const assert = require('assert');
const { describe, it } = require('node:test');
describe('Array', function() {
    describe('#indexOf()', () => {
        it('should return -1 when the value is not present', function() {
            const arr = [1, 2, 3];
            assert.strictEqual(arr.indexOf(4), -1);
        });
    });
});
```

### Example 2: Async Test Case

**Before:**

```js
const assert = require('assert');
describe('Async Test', function() {
    it('should complete after a delay', async function(done) {
        const result = await new Promise(resolve => setTimeout(() => resolve(42), 100));
        assert.strictEqual(result, 42);
    });
});
```

**After:**

```js
const assert = require('assert');
const { describe, it } = require('node:test');
describe('Async Test', () => {
    it('should complete after a delay', async () => {
        const result = await new Promise(resolve => setTimeout(() => resolve(42), 100));
        assert.strictEqual(result, 42);
    });
});
```

### Example 3: Hooks

**Before:**

```js
const assert = require('assert');
const fs = require('fs');
describe('File System', function() {
    before(function() {
        fs.writeFileSync('test.txt', 'Hello, World!');
    });

    after(function() {
        fs.unlinkSync('test.txt');
    });

    it('should read the file', function() {
        const content = fs.readFileSync('test.txt', 'utf8');
        assert.strictEqual(content, 'Hello, World!');
    });
});
```

**After:**

```js
const assert = require('assert');
const fs = require('fs');
const { describe, before, after, it } = require('node:test');
describe('File System', () => {
    before(() => {
        fs.writeFileSync('test.txt', 'Hello, World!');
    });

    after(() => {
        fs.unlinkSync('test.txt');
    });

    it('should read the file', () => {
        const content = fs.readFileSync('test.txt', 'utf8');
        assert.strictEqual(content, 'Hello, World!');
    });
});
```

### Example 4: `done` Callback

**Before:**

```js
const assert = require('assert');
describe('Callback Test', function() {
    it('should call done when complete', function(done) {
        setTimeout(() => {
            assert.strictEqual(1 + 1, 2);
            done();
        }, 100);
    });
});
```

**After:**

```js
const assert = require('assert');
const { describe, it } = require('node:test');
describe('Callback Test', () => {
    it('should call done when complete', (t, done) => {
        setTimeout(() => {
            assert.strictEqual(1 + 1, 2);
            done();
        }, 100);
    });
});
```

### Example 5: Skipped Tests

**Before:**

```js
const assert = require('assert');
describe('Skipped Test', function() {
    it.skip('should not run this test', function() {
        assert.strictEqual(1 + 1, 3);
    });
    it('should also be skipped', function() {
        this.skip();
        assert.strictEqual(1 + 1, 3);
    });
});
```

**After:**

```js
const assert = require('assert');
const { describe, it } = require('node:test');
describe('Skipped Test', () => {
    it.skip('should not run this test', () => {
        assert.strictEqual(1 + 1, 3);
    });
    it('should also be skipped', (t) => {
        assert.strictEqual(1 + 1, 3);
        t.skip();
    });
});
```

### Example 6: Dynamic/generated tests

**Before:**

```js
const assert = require('assert');
describe('Dynamic Tests', function() {
    const tests = [1, 2, 3];
    tests.forEach((test) => {
        it(`should handle test ${test}`, function() {
            assert.strictEqual(test % 2, 0);
        });
    });
});
```

**After:**

```js
const assert = require('assert');
const { describe, it } = require('node:test');
describe('Dynamic Tests', () => {
    const tests = [1, 2, 3];
    tests.forEach((test) => {
        it(`should handle test ${test}`, () => {
            assert.strictEqual(test % 2, 0);
        });
    });
});
```

### Example 7: Timeouts handling

This timeout handling work for `describe`, `it` and hooks.

**Before:**

```js
const assert = require('assert');
describe('Timeout Test', function() {
    this.timeout(500);

    it('should complete within 100ms', function(done) {
        this.timeout(100);
        setTimeout(() => {}, 500); // This will fail

    });

    it('should complete within 200ms', function() {
        this.timeout(200);
        setTimeout(() => {}, 100); // This will pass
    });
});
```

**After:**

> [!NOTE]
> I can't make the example below work. 🥲
> _in theory, it should work_

```js
const assert = require('assert');
const { describe, it } = require('node:test');
describe('Timeout Test', { timeout: 500 }, () => {
    it('should complete within 100ms', { timeout: 100 }, () => {
        setTimeout(() => {}, 500); // This will fail
    });

    it('should complete within 200ms', { timeout: 200 }, () => {
        setTimeout(() => {}, 100); // This will pass
    });
});
```

## Caveats

- `node:test` doesn't support the `retry` option that Mocha has, so any tests using that will need to be handled separately.

## REFS

- [Node Test Runner](https://nodejs.org/api/test.html)
- [Mocha](https://mochajs.org/)

Xstoudi:
Should fix #103


recipes/mocha-to-node-test-runner/README.md
```diff
@@ -0,0 +1,153 @@
...
+                setTimeout(() => {
+                        assert.strictEqual(1 + 1, 2);
+                        done();
+                }, 100);
+        });
+})
+```
+
+### Example 5: Skipped
+```diff
```
AugustinMauroy:
````suggestion

```diff
```

recipes/mocha-to-node-test-runner/README.md
```diff
@@ -0,0 +1,153 @@
...
+-       it('should complete within 200ms', function(done) {
+-               this.timeout(200);
++       it('should complete within 200ms', { timeout: 200 }, function(t, done) {
++
+              setTimeout(done, 100); // This will pass
+      });
+});
+```
+## Caveats
+* `node:test` doesn't support the `retry` option that Mocha has, so any tests using that will need to be handled separately.
```
AugustinMauroy:
````suggestion
```

## Caveats

* `node:test` doesn't support the `retry` option that Mocha has, so any tests using that will need to be handled separately.
```

File: recipes/mocha-to-node-test-runner/README.md [hunk 3]
```diff
@@ -0,0 +1,153 @@
+# Mocha to Node.js Test Runner
```
AugustinMauroy:
```suggestion
# Mocha to Node.js Test Runner

```

recipes/mocha-to-node-test-runner/src/workflow.ts
```diff
@@ -0,0 +1,246 @@
...
+
+	const usedGlobalIdentifiers = globalIdentifiers.filter((globalIdentifier) =>
+		['', '.skip', '.only']
+			.map((suffix) => `${globalIdentifier}${suffix}($$$)`)
+			.some((pattern) => rootNode.findAll({ rule: { pattern } }).length > 0),
+	);
+
+	if (usedGlobalIdentifiers.length === 0) {
+		return null;
+	}
```
AugustinMauroy:
```suggestion
	if (!usedGlobalIdentifiers.length) return null;
```

recipes/mocha-to-node-test-runner/src/workflow.ts
```diff
@@ -0,0 +1,246 @@
...
+	].flatMap((transform) => transform(root));
+	if (!edits.length) {
+		return null;
+	}
+
+	return rootNode
+		.commitEdits(edits)
+		.split('\n')
+		.map((line) => (line.trim() === '' ? line.trim() : line))
+		.join('\n');
```
AugustinMauroy: what the goal of that ?

utils/src/is-esm.ts
AugustinMauroy: Can we have unit test here ?

AugustinMauroy (commented):
Oh nice that really cool !

Few points that can be improved:
- md format on readme 
- little bit of JSdoc 
- small missing unit test

recipes/mocha-to-node-test-runner/src/workflow.ts
Xstoudi: When removing `this.timeout(x)` instructions, the indent remains as it's not in the AST.
Those three lines remove whitespaces in lines that only contains whitespaces.

Not sure if there is already an util to achieve it. It seems like fully removing line is still a bit painful using codemod.

utils/src/is-esm.ts
Xstoudi: Done

recipes/mocha-to-node-test-runner/src/workflow.ts
```diff
@@ -0,0 +1,246 @@
...
+	if (usedGlobalIdentifiers.length === 0) {
+		return null;
+	}
+
+	const edits = [
+		transformImport,
+		transformDoneCallbacks,
+		transformThisSkip,
+		transformThisTimeout,
+	].flatMap((transform) => transform(root));
```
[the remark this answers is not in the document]
Xstoudi: Multiple calls, then have to merge all resulting Edit[] arrays... that's a flatmap! 😀

recipes/mocha-to-node-test-runner/src/workflow.ts
AugustinMauroy: We have an utility for that on utility package called `removeLines` which allow you to pass array of `Range` .

So you can use it at:

1. intanciate an array
2. when editing pushing on this array range of thing you want to remove
2. after pushing "commit" call the function and return his result

recipes/mocha-to-node-test-runner/tests/input/1_basic.js
```diff
@@ -0,0 +1,10 @@
+const assert = require('assert');
+
+describe('Array', function() {
+	describe.skip('#indexOf()', function() {
+		it('should return -1 when the value is not present', function() {
+			const arr = [1, 2, 3];
+			assert.strictEqual(arr.indexOf(4), -1);
+		});
```
AugustinMauroy: maybe add an other `it` that use arrow function as call back 
and a fonction as arg like
```js
function myBasicTest = () => {
  assert(true)
}
```

The second case will be can super complicated to migrate. 
in a first time let's just have test that proof that's ins't handled and document it.
In second time (follow up pr) having systems that handle this case. The logic to catch a refercen is already furnished by the codemod context https://docs.codemod.com/jssg/semantic-analysis

recipes/mocha-to-node-test-runner/src/workflow.ts
Xstoudi: Doing so still leads to troubles.
- transform A edit lines (eg. 2, 6, 10)
- transform B add one more line between 3 and 4
- apply edits
- remove lines 2, 6, 10 -> oopsie, 6 and 10 are deleted instead of 7 and 11 instead.
For `mocha-to-node-test-runner` : it's gonna happen, no doubt.

I'm gonna try to implement an offset tracking for removed lines, but this is gonna bloat the code and imho we should let the formatter format things.

Xstoudi:
Let me know if anything more is required.

AugustinMauroy (commented):
LGMT ! for the code transformation part missing the removing dep you should take code from `chalk-to-styletext`.

Xstoudi:
Added that, let me know if something is missing.

AugustinMauroy (approved):
For me it's good ! we need to wait a second approval to land !

AugustinMauroy:
F*ck what happened here. !Windows!

cc @nodejs/userland-migrations any idea

AugustinMauroy (approved):
LGTM ! I stil don't get why windows give this error
Maybe use lower comparison level on testing https://docs.codemod.com/jssg/testing#strictness-levels

brunocroh (approved):
LGTM

recipes/mocha-to-node-test-runner/src/workflow.ts
```diff
@@ -0,0 +1,255 @@
...
+			});
+			const thisKeyword = memberExpr?.field('object');
+			if (thisKeyword) {
+				edits.push(thisKeyword.replace('t'));
+			}
+
+			const fn = findEnclosingFunction(call);
+			if (!fn) {
+				return edits;
+			}
```
JakobJingleheimer: same nit


```suggestion
			if (!fn) return edits;
```

recipes/mocha-to-node-test-runner/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+This recipe migrate Mocha v8 tests to Node.js test runner (v22, v24+)
```
JakobJingleheimer:
```suggestion
This migrates Mocha 8.x tests to Node.js (22.x, 24.x) test runner
```

recipes/mocha-to-node-test-runner/README.md
```diff
@@ -0,0 +1,171 @@
...
++const { describe, it } = require('node:test');
+
+ describe('Array', function() {
+        describe.skip('#indexOf()', function() {
+                it('should return -1 when the value is not present', function() {
+                        const arr = [1, 2, 3];
+                        assert.strictEqual(arr.indexOf(4), -1);
+                });
+        });
+ });
```
JakobJingleheimer: nit: spaces (instead of tabs) please in markdown codeblocks because most viewers render tabs enormous 😭

utils/src/is-esm.ts
```diff
@@ -0,0 +1,45 @@
...
+				field: 'function',
+				regex: 'require',
+			},
+		},
+	});
+	if (usingRequire) {
+		return false;
+	}
+
+	const packageJsonPath = join(process.cwd(), 'package.json');
```
JakobJingleheimer: @AugustinMauroy what was the verdict on [`node:module.findPackageJSON()`](https://nodejs.org/api/module.html#modulefindpackagejsonspecifier-base)?

JakobJingleheimer (commented):
Thanks for this! 🙌

A few nits here and there.

For the test files, could the names please be more descriptive (eg `3_hooks` is not so illuminating).

Also, there are 2 options for organising test files. We do have a bit of a mix, but [option 1](https://docs.codemod.com/jssg/testing#single-file-fixtures) is far superior to [the "junk-drawer" option 2](https://docs.codemod.com/jssg/testing#directory-snapshot-fixtures). Would you mind adjusting these to option 1? If you don't want to bother with it, that's fine—but would you mind if one of us made the adjustment within this PR?

I haven't reviewed the test cases yet because it's so cumbersome with option 2.

AugustinMauroy (approved):

brunocroh (approved):

brunocroh (approved):

brunocroh (approved):

PR #314, opened 2025-12-22
ci: publish codemods as trusted publisher

### LINKED ISSUE #307: setup(`publish`): use trusted published
Codemod now supports publishing via GitHub Actions OIDC. No API keys required.
If your workflow runs in a repo under the nodejs org, you can publish codemods under the @nodejs scope with zero configuration.
https://github.com/codemod/publish-action
Example workflow:
https://github.com/codemod/cypress-to-playwright/blob/main/.github/workflows/release-on-tag.yml
For more info: [here is the docs](https://docs.codemod.com/publishing#trusted-publishers)

What do you think

Xstoudi:
Opening this PR as an implementation example of @AugustinMauroy proposal in #307.

If merged, you'll probably want to delete the `CODEMOD_TOKEN` secret.
No other configuration should be required as `nodejs` org on GitHub is also `nodejs` org on Codemod.

Fixes #307 if merged

### DIFF
--- a/.github/workflows/codemod_publish.yml
+++ b/.github/workflows/codemod_publish.yml
@@ -15,15 +15,6 @@ on:
         required: true
         type: string
 
-permissions:
-    id-token: write
-    contents: read
-
-jobs:
-  validate-and-publish:
-    name: Validate and Publish Codemod
-    runs-on: ubuntu-latest
-
 permissions: read-all
 
 jobs:
@@ -32,8 +23,8 @@ jobs:
     runs-on: ubuntu-latest
 
     permissions:
-        id-token: write
-        contents: read
+      id-token: write
+      contents: read
     outputs:
       version: ${{ steps.parse-tag.outputs.version }}
       codemod-name: ${{ steps.parse-tag.outputs.codemod-name }}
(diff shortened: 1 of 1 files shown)


mohebifar:
> first publish need to be done from local env like Jacob or me

That won't not needed if you use OIDC publishers. As long as the github actions is run in a repo in the `nodejs` org matching the `@nodejs` scope in codemod.

> then link the codemod on registry to the repo

That again is not needed as long as the codemods are scoped `@nodejs/`.

JakobJingleheimer (dismissed):

.github/workflows/codemod_publish.yml
```diff
@@ -103,14 +107,11 @@ jobs:
...
-      - name: Authenticate with Codemod registry
-        env:
-          CODEMOD_TOKEN: ${{ secrets.CODEMOD_TOKEN }}
-        run: npx codemod login --api-key "$CODEMOD_TOKEN"
 
-      - name: Publish codemod
-        working-directory: ${{ steps.parse-tag.outputs.codemod-path }}
-        run: npx codemod publish
+      - name: Publish codemod to registry
+        uses: codemod/publish-action@dd6c8dbc5ceb1a6146feba41481d88b43da50024
```
avivkeller: Add the version at the end

```suggestion
        uses: codemod/publish-action@dd6c8dbc5ceb1a6146feba41481d88b43da50024 # v1
```

avivkeller (commented):
Define the permissions _inside_ the job.

PR #326, opened 2026-01-03
Feat: add getModuleDependencies utillity

brunocroh:
Add a getModuleDependencies utility to replace cases like the one below:

```ts
  return [
		...getNodeRequireCalls(node, name),
		...getNodeImportStatements(node, name),
		...getNodeImportCalls(node, name),
	];
```

Also update the recipes to use it.

### DIFF
--- a/utils/README.md
+++ b/utils/README.md
@@ -17,6 +17,21 @@ These utilities provide battle-tested solutions for common codemod operations, r
 
 ### Import and Require Detection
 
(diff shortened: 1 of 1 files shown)


AugustinMauroy:
Noice !

could you document it in the utils readme ?

AugustinMauroy (approved):

JakobJingleheimer:
Yep, sounds good 🙂

PR #340, opened 2026-01-21
feat(`cjs-to-esm`): scaffold codemod 

AugustinMauroy:
## Description

After some research on this plus spec written by @JakobJingleheimer. I'm marking this pr as ready to review.
So what does it do:
- Setup the codemod structure 
- setupping the data sharing part 
- setuping tests without file and without running them

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
       latest: ${{ steps.set-matrix.outputs.requireds }}
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
+        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
         with:
           egress-policy: audit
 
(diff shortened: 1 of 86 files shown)


recipes/cjs-to-esm/README.md
```diff
@@ -0,0 +1,15 @@
...
+# ESM Migration Codemod
+
+## Context-local Variable Migration
+
+This codemod includes a step to help migrate context-local variables and Node.js built-in globals from CommonJS to ESM. It detects usages of:
+
+- `__filename` → `import.meta.filename` (Node.js v20.11.0 / v21.2.0)
+- `__dirname` → `import.meta.dirname` (Node.js v20.11.0 / v21.2.0)
+- `require.main` → `import.meta.main` (Node.js v22.18.0 / v24.2.0)
+- `require.resolve` → `import.meta.resolve` (available)
```
avivkeller: These variables behave a bit differently. Wouldn't it be more one-to-one to use `createRequire().resolve`?

JakobJingleheimer:
I think it would be better to use a feature branch as a staging area (and then merge that into `main` once it's ready for v`0.1.0`); I would name that branch `cjs-to-esm` (perhaps rename this one to something like `cjs-to-esm/scaffold`).

Aside from that, sure this is fine. I think we generally don't need to do this sort of ultra-basic PR that just stubs things 😅

AugustinMauroy:
> I think it would be better to use a feature branch as a staging area (and then merge that into `main` once it's ready for v`0.1.0`); I would name that branch `cjs-to-esm` (perhaps rename this one to something like `cjs-to-esm/scaffold`).

Done and make sens 

> Aside from that, sure this is fine. I think we generally don't need to do this sort of ultra-basic PR that just stubs things 😅

you was asking for small pr I do small pr 😎

JakobJingleheimer (approved):
🙌

PR #382, opened 2026-02-25
fix(fs-access-mode-constants): don't modify other thing

AugustinMauroy:
## Description 

Alex (codemod) reported that this codemod make changes on other file. 

So:
- I change tests file name for better understanding 
- I refracto to use more utility in goal of reducing complexity


JakobJingleheimer (commented):
Thanks for this!

How-so does the migration unintentionally affect other files?

JakobJingleheimer (commented):
New changes look good 🙂

AugustinMauroy:
@JakobJingleheimer added

brunocroh (approved):

recipes/fs-access-mode-constants/src/workflow.ts
```diff
@@ -259,6 +265,6 @@ function rewriteCollectedBindings({
 	};
 }
 
-function escapeRegExp(text: string): string {
+export function escapeRegExp(text: string): string {
 	return text.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
```
JakobJingleheimer: Is this a homebrewed utility for [`RegExp.escape()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape)?

JakobJingleheimer (approved):

recipes/fs-access-mode-constants/src/workflow.ts
AugustinMauroy: ast-grep is rust like regex so if I'm not doing wrong this regex catch regex rust like syntax

brunocroh (approved):

PR #413, opened 2026-03-29
feat(DEP0164): handle non-integer codes passed to `process.exit` & `process.exitCode`

KevinSailema:
## Summary
This PR adds a new codemod recipe to handle DEP0164 by coercing unsupported values passed to process.exit(code) and assigned to process.exitCode into integer-compatible values.

Closes #405.

## What this adds
- New recipe: @nodejs/process-exit-coercion-to-integer
- New transform to migrate:
  - process.exit(value)
  - process.exitCode = value
- Full recipe wiring:
  - codemod manifest
  - workflow definition
  - package configuration
  - input/expected fixture tests

## Transformation behavior
The codemod applies explicit and safe coercions aligned with the issue requirements:

- Preserves valid values (no change):
  - undefined
  - null
  - integer numbers
  - integer strings (example: "1")
- Converts boolean values:
  - process.exit(true) -> process.exit(1)
  - process.exit(false) -> process.exit(0)
  - process.exitCode = success -> process.exitCode = success ? 0 : 1
- Converts floating-point and numeric expressions explicitly:
  - process.exit(1.5) -> process.exit(Math.floor(1.5))
  - process.exit(0.5 + 0.7) -> process.exit(Math.floor(0.5 + 0.7))
- Converts non-integer string literals in process.exit to 1
- Handles object literal assignment for process.exitCode with code extraction when present:
  - process.exitCode = { code: 1 } -> process.exitCode = 1

## Safety and scope
- Conservative by design: ambiguous identifier values are left unchanged unless safely inferable from local literal initialization.
- No dynamic evaluation and no runtime behavior injection.
- No unrelated refactors or formatting-only changes.

## Tests
Added fixture coverage for the 10 scenarios from the issue plus additional edge cases:
- import/require aliases for process
- unknown identifier no-op behavior
- object code field with float coercion

Validation run:
- Recipe tests passed
- Repository pre-commit checks passed (lint, type-check, workspace tests)


AugustinMauroy (changes requested):
Also missing readme.

AugustinMauroy (changes requested):
Better but still something to improve

AugustinMauroy (approved):
LGMT !

recipes/process-exit-coercion-to-integer/src/workflow.ts
```diff
@@ -0,0 +1,274 @@
...
+}
+
+function floorWrap(expressionText: string): string {
+	return `Math.floor(${expressionText})`;
+}
+
+function coerceBoolean(expressionText: string, mode: ExitMode): string {
+	if (mode === 'exit') return `${expressionText} ? 1 : 0`;
+	return `${expressionText} ? 0 : 1`;
+}
```
brunocroh, line 80: I think we can just coerce booleans with `Number`to be more simple and consistent.

```diff
- process.exit(true)
+ process.exit(Number(true))
```

recipes/process-exit-coercion-to-integer/src/workflow.ts
```diff
@@ -0,0 +1,274 @@
...
+	return null;
+}
+
+function coerceFromObjectLiteral(
+	objectNode: SgNode<JS>,
+	mode: ExitMode,
+): string {
+	if (mode !== 'exitCode') return '1';
+
+	const codeValue = getObjectCodeValue(objectNode);
```
brunocroh, line 105: inline it

```suggestion
		const pair = objectNode.find({
		rule: {
			kind: 'pair',
			has: {
				kind: 'property_identifier',
				field: 'key',
				regex: 'code',
			},
		},
	});

	const valueValue = pair?.field('value') || null;
```

recipes/process-exit-coercion-to-integer/src/workflow.ts
```diff
@@ -0,0 +1,274 @@
...
+
+	for (const pair of pairs) {
+		const key = pair.field('key');
+		const value = pair.field('value');
+		if (!key || !value) continue;
+		if (key.text() === 'code') return value;
+	}
+
+	return null;
+}
```
brunocroh, line 97: it can be simplified, and doing that I think just make it inline is better, so replace the functions/implementation by the suggestion on the next comment
```suggestion
```

recipes/process-exit-coercion-to-integer/src/workflow.ts
```diff
@@ -0,0 +1,274 @@
...
+
+	if (inferredKind === 'boolean_true' || inferredKind === 'boolean_false') {
+		return mode === 'exit'
+			? inferredKind === 'boolean_true'
+				? '1'
+				: '0'
+			: inferredKind === 'boolean_true'
+				? '0'
+				: '1';
+	}
```
brunocroh, line 158: why `inferredKind` need to be compared with same value so many times? this code is quite confusing to me, could you explain why it is necessary or maybe refactor it a bit?

recipes/process-exit-coercion-to-integer/src/workflow.ts
```diff
@@ -0,0 +1,274 @@
...
+
+function collectInferredIdentifiers(
+	rootNode: SgNode<JS>,
+): Map<string, InferredIdentifier> {
+	const inferred = new Map<string, InferredIdentifier>();
+	const declarators = rootNode.findAll({
+		rule: {
+			kind: 'variable_declarator',
+		},
+	});
```
brunocroh, line 196: this will look for all variable_declarators in the source code I think it not really needed.

https://docs.codemod.com/jssg/semantic-analysis#node-definition-options

LSP.goToDefinition is supported by jssg try use it instead

recipes/process-exit-coercion-to-integer/tests/11-import-require-aliases/expected.js
```diff
@@ -0,0 +1,6 @@
+import proc from 'node:process';
+const { exit } = require('node:process');
+
+proc.exit(0);
+exit(Math.floor(2.3));
+proc.exitCode = 0;
```
brunocroh, line 6: Why `process.exitCode = true` becomes `process.exitCode = 0`? It not need to be `1` instead?

brunocroh (changes requested):

AugustinMauroy:
bump @KevinSailema

PR #432, opened 2026-04-17
fix(tls-create-secure-pair-to-tls-socket): ignore coments when parsing args

brunocroh:
In situations where there are inline comments, it parses the function arguments incorrectly. Now it ignores the comments, and a test scenario has been added.

```ts
const securePair = tls.createSecurePair(
	tls.createSecureContext(options),
	true,   // isServer
	false,  // requestCert
	false   // rejectUnauthorized
)
```


AugustinMauroy (commented):
Nice catch, but need bump package/codemod version so we can publish it

brunocroh:
> Nice catch, but need bump package/codemod version so we can publish it

Whoops, sorted.

AugustinMauroy (approved):

JakobJingleheimer (approved):
🙌 Nice!

The new tests are using the wrong organisation though. Otherwise g2g (approving presumptively so it's not a blocker once fixed).

JakobJingleheimer (approved):

PR #444, opened 2026-04-26
setup(workflow): narrow `CI` path filtering

JakobJingleheimer:
The `CI` workflow is concerned with implementation code (basically just typescript). We don't need to run linting, type-checking, and tests when non-code files change (ex markdown or yaml), and the yaml lint job is in a dedicated workflow now, so the `CI` job can filter just on js/ts and package.json files (where npm commands can change, affecting CI tasks).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
       - "recipes/**/*.ts"
       - "recipes/**/package.json"
       - "utils/**/*.ts"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
AugustinMauroy, line 1: sould also catch js file so if we update test fixtures it's will trigger the ci

AugustinMauroy (approved):

brunocroh (approved):

PR #445, opened 2026-04-26
setup(workflow): run `Code Quality` → `tests` only on changed workspace(s)

JakobJingleheimer:
Wait for https://github.com/nodejs/userland-migrations/pull/444

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,7 +115,9 @@ jobs:
 
       - run: npm ci
 
-      - run: >
+      - name: Run tests (determined by changeset)
+        shell: bash
+        run: >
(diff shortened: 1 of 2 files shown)


AugustinMauroy:
BTW seem good but if utils package is touch we should run all jssg tests to be sure we don't break anything by side effect

JakobJingleheimer:
> BTW seem good but if utils package is touch we should run all jssg tests to be sure we don't break anything by side effect

Ah, true!

AugustinMauroy (dismissed):

AugustinMauroy (approved):

JakobJingleheimer:
Okay, done with this now @AugustinMauroy (sorry for the spam).

brunocroh (approved):

PR #451, opened 2026-04-30
fix(workflow): support dependabot pr titles

brunocroh:
The PRs created by Dependabot do not have valid titles for our CI, so I updated the regex to support dep: bump ... as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,7 @@ updates:
 
   - package-ecosystem: github-actions
     commit-message:
-      prefix: dep(github-actions)
+      prefix: dep(GHA)
     directory: /
     labels:
       - "dependencies"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,7 @@ updates:
 
   - package-ecosystem: github-actions
     commit-message:
-      prefix: dep
+      prefix: dep(github-actions)
     directory: /
     labels:
       - "dependencies"
@@ -21,7 +21,8 @@ updates:
 
   - package-ecosystem: npm # See documentation for possible values
     commit-message:
-      prefix: dep
+      prefix: dep(prod)
+      prefix-development: dep(dev)
     directories:
       - / # Location of package manifests
       - /recipies/*
--- a/.github/scripts/pr-title.ts
+++ b/.github/scripts/pr-title.ts
@@ -4,8 +4,8 @@ import { env } from 'node:process';
 import { SCOPE_RGX, SUPPORTED_PREFIXES } from './pr-prefixes.ts';
 
 declare global {
-  namespace NodeJS {
-    interface ProcessEnv {
+	namespace NodeJS {
+		interface ProcessEnv {
 			PR_TITLE: string;
 		}
 	}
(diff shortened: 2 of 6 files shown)


.github/scripts/pr-title.ts
```diff
@@ -4,22 +4,22 @@ import { env } from 'node:process';
...
-const INVALID_PR_TITLE_ERR = '\
+const INVALID_PR_TITLE_ERR =
+	'\
 The pull request title did not match the required format; see CONTRIBUTING.md for details.\
 ';
 
 assert.match(PR_TITLE, SCOPE_RGX, INVALID_PR_TITLE_ERR);
-const firstParen = PR_TITLE.indexOf('(');
-const prefix = PR_TITLE.slice(0, firstParen);
+const [_, prefix] = SCOPE_RGX.exec(PR_TITLE) || [];
```
brunocroh: I added an alternative in the regex and captured the prefix for both scenarios. I think this is simpler than implementing a separate approach here, especially since this PR introduces `dep: bump` as a valid case as well, which doesn’t have parentheses.

But if this isn't good for some reason, let me know

AugustinMauroy (approved):
LGMT !

JakobJingleheimer:
I think it would be better to update Dependabot's config to include scope (and then the PR title will pass).

brunocroh:
> I think it would be better to update Dependabot's config to include scope (and then the PR title will pass).

Oh yeah, I agree—that sounds better. What should the prefix be?

dep(workspace)? Or is there a clearer name for the scope?

JakobJingleheimer:
> dep(workspace)? Or is there a clearer name for the scope?

~Yeah that sounds good for the gh action ones~

Edit: Wait, do you mean `dep(workflow)`?

There are 2 dependapot categories defined in its config (the other is for npm). In the npm case, if possible something like `dep(dev)` and `dep(prod)`? (IIR, dependabot can differentiate dev vs prod)

brunocroh:
> > dep(workspace)? Or is there a clearer name for the scope?
> 
> ~Yeah that sounds good for the gh action ones~
> 
> Edit: Wait, do you mean `dep(workflow)`?
> 
> There are 2 dependapot categories defined in its config (the other is for npm). In the npm case, if possible something like `dep(dev)` and `dep(prod)`? (IIR, dependabot can differentiate dev vs prod)

What do dev and prod actually represent in our case?
I was thinking of using dep(workspace) for both, but it’s probably better to distinguish between them, e.g.:
```
  - package-ecosystem: github-actions
    commit-message:
      prefix: dep(gh-actions)

  - package-ecosystem: npm
    commit-message:
      prefix: dep(npm)

```

JakobJingleheimer:
> What do dev and prod actually represent in our case?

An npm development dependency  (eg `"devDependencies"`) and an npm production dependency (eg `"dependencies"`)

brunocroh:
> > What do dev and prod actually represent in our case?
> 
> An npm development dependency (eg `"devDependencies"`) and an npm production dependency (eg `"dependencies"`)

Got it, done. Let me know if I missed anything or if there’s any room for improvement. Thanks for your patience 😁

.github/dependabot.yml
```diff
@@ -11,7 +11,7 @@ updates:
 
   - package-ecosystem: github-actions
     commit-message:
-      prefix: dep
+      prefix: dep(github-actions)
```
JakobJingleheimer: Nit: this is a bit long for something included in an already-long title

```suggestion
      prefix: dep(GHA)
```


```suggestion
      prefix: dep(actions)
```

JakobJingleheimer (approved):
🙌 ha, no patience needed. Thanks for handling!

JakobJingleheimer (approved):

PR #461, opened 2026-05-08
fix(codemod): broken bundle

AugustinMauroy:
I see that there are package with low level of running so I tried to know if there are missing case and just realized that they fail for broken bundle.


brunocroh (approved):

brunocroh:
good catch mate

brunocroh (approved):

JakobJingleheimer (approved):

PR #473, opened 2026-05-24
feat(DEP0153): add dns lookup options codemod

Herrtian:
Refs #409.

Adds a DEP0153 recipe for literal `dns.lookup()` and `dnsPromises.lookup()` option values. Dynamic option values are left unchanged for manual review.

Checked with:
- `npm run test --workspace @nodejs/dns-lookup-options-coercion`
- `npm run type-check`
- `npm run lint`
- `git diff --check`

### DIFF
--- a/recipes/dns-lookup-options-coercion/src/workflow.ts
+++ b/recipes/dns-lookup-options-coercion/src/workflow.ts
@@ -12,19 +12,26 @@ export default function transform(root: SgRoot<Js>): string | null {
 	const edits: Edit[] = [];
 	const lookupCallees = collectLookupCallees(root);
 
-	if (lookupCallees.size === 0) return null;
+	if (!lookupCallees.size) return null;
 
 	for (const callee of lookupCallees) {
-		const calls = rootNode.findAll({
-			rule: { pattern: `${callee}($$$_ARGS)` },
+		const calls = rootNode.findAll<'call_expression'>({
+			rule: {
+				kind: 'call_expression',
+				has: {
+					field: 'function',
+					kind: callee.includes('.') ? 'member_expression' : 'identifier',
+					pattern: callee,
+				},
+			},
 		});
 
 		for (const call of calls) {
-			edits.push(...transformLookupCall(call));
+			processLookupCall(call, edits);
 		}
 	}
 
-	if (edits.length === 0) return null;
+	if (!edits.length) return null;
 
 	return rootNode.commitEdits(edits);
 }
@@ -62,17 +69,17 @@ function addResolvedBinding(
  * DEP0153 affects option values, so calls without options or with a shared
  * options variable are intentionally left unchanged for manual review.
  */
-function transformLookupCall(call: SgNode<Js>): Edit[] {
+function processLookupCall(call: SgNode<Js>, edits: Edit[]): void {
 	const args = call.field('arguments');
-	if (!args) return [];
+	if (!args) return;
 
 	const optionArg = args
 		.children()
 		.filter((child) => !IGNORED_ARGUMENT_KINDS.has(child.kind()))[1];
 
-	if (!optionArg || optionArg.kind() !== 'object') return [];
+	if (!optionArg || optionArg.kind() !== 'object') return;
 
-	return transformOptionsObject(optionArg);
+	edits.push(...transformOptionsObject(optionArg));
 }
 
 /**
@@ -82,7 +89,7 @@ function transformLookupCall(call: SgNode<Js>): Edit[] {
  */
 function transformOptionsObject(options: SgNode<Js>): Edit[] {
 	const edits: Edit[] = [];
-	const pairs = options.children().filter((child) => child.kind() === 'pair');
+	const pairs = options.findAll<'pair'>({ rule: { kind: 'pair' } });
 
 	for (const pair of pairs) {
 		const key = getOptionKey(pair);
@@ -100,7 +107,7 @@ function transformOptionsObject(options: SgNode<Js>): Edit[] {
 	return edits;
 }
 
-function getOptionKey(pair: SgNode<Js>): string | null {
+function getOptionKey(pair: SgNode<Js, 'pair'>): string | null {
 	const key = pair.field('key');
 	if (!key) return null;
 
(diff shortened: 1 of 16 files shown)


recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,153 @@
...
+export default function transform(root: SgRoot<Js>): string | null {
+	const rootNode = root.root();
+	const edits: Edit[] = [];
+	const lookupCallees = collectLookupCallees(root);
+
+	if (lookupCallees.size === 0) return null;
+
+	for (const callee of lookupCallees) {
+		const calls = rootNode.findAll({
+			rule: { pattern: `${callee}($$$_ARGS)` },
```
AugustinMauroy: use better ast based query like using king ecxt

recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,153 @@
...
+		});
+
+		for (const call of calls) {
+			edits.push(...transformLookupCall(call));
```
AugustinMauroy:
```suggestion
			edits.push(transformLookupCall(call));
```

why do we need to spread ?

recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,153 @@
...
+const NUMERIC_OPTIONS = new Set(['family', 'hints']);
+const BOOLEAN_OPTIONS = new Set(['all', 'verbatim']);
+const IGNORED_ARGUMENT_KINDS = new Set([',', '(', ')']);
+
+export default function transform(root: SgRoot<Js>): string | null {
+	const rootNode = root.root();
+	const edits: Edit[] = [];
+	const lookupCallees = collectLookupCallees(root);
+
+	if (lookupCallees.size === 0) return null;
```
AugustinMauroy:
```suggestion
	if (!lookupCallees.size) return null;
```

recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,153 @@
...
+		const calls = rootNode.findAll({
+			rule: { pattern: `${callee}($$$_ARGS)` },
+		});
+
+		for (const call of calls) {
+			edits.push(...transformLookupCall(call));
+		}
+	}
+
+	if (edits.length === 0) return null;
```
AugustinMauroy:
```suggestion
	if (!edits.length) return null;
```

recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,153 @@
...
+
+		if (replacement !== null) {
+			edits.push(value.replace(replacement));
+		}
+	}
+
+	return edits;
+}
+
+function getOptionKey(pair: SgNode<Js>): string | null {
```
AugustinMauroy:
```suggestion
function getOptionKey(pair: SgNode<Js, 'pair'>): string | null {
```

FYI: you can give kind of node to have better typing

recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,153 @@
...
+		.filter((child) => !IGNORED_ARGUMENT_KINDS.has(child.kind()))[1];
+
+	if (!optionArg || optionArg.kind() !== 'object') return [];
+
+	return transformOptionsObject(optionArg);
+}
+
+function transformOptionsObject(options: SgNode<Js>): Edit[] {
+	const edits: Edit[] = [];
+	const pairs = options.children().filter((child) => child.kind() === 'pair');
```
AugustinMauroy:
```suggestion
	const pairs = options.children().filter((child) => child.is('pair'));
```

never use `kind() === 'foo'` when you can use the `is` method which is designed for that

recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,153 @@
...
+	const key = pair.field('key');
+	if (!key) return null;
+
+	if (key.kind() === 'property_identifier') {
+		return key.text();
+	}
+
+	if (key.kind() === 'string') {
+		return getStringLiteralValue(key);
+	}
```
AugustinMauroy:
```suggestion
	const keyKind = key.kind();
	
	if (keyKind === 'property_identifier') {
		return key.text();
	}

	if (keyKind === 'string') {
		return getStringLiteralValue(key);
	}
```

here since you use 2 times the kind it's better to use `===`

recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+function getNumericReplacement(value: SgNode<Js>): string | null {
+	if (value.kind() !== 'string') return null;
+
+	const stringValue = getStringLiteralValue(value);
+	if (!stringValue || !/^\d+$/.test(stringValue)) return null;
+
+	return String(Number.parseInt(stringValue, 10));
+}
```
AugustinMauroy, line 155: this seem over engineered.

AugustinMauroy (commented):
Nice work ! But cold you add more jsDocs around sub function to know why/when/how is it useful

test suite seem good bases on issue 

in readme you said

> This recipe only changes literal option values. Dynamic values such as `{ family: familyOption }` need manual review.

what if we use semantic analitic you can found where the option is defined and maybe update it https://docs.codemod.com/jssg/semantic-analysis#api-reference

Herrtian:
Added JSDoc around the option rewrite path and noted why identifier propagation is left for manual review.

Checked with:
- npm run test --workspace @nodejs/dns-lookup-options-coercion
- npm run type-check
- npm run lint\n- git diff --check

recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,171 @@
...
+
+function getBooleanReplacement(value: SgNode<Js>): string | null {
+	if (value.kind() === 'number') {
```
AugustinMauroy: here need to cache kind value and then use `switch`

recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,171 @@
...
+		if (value.text() === '1') return 'true';
+	}
+
+	if (value.kind() !== 'string') return null;
+
+	const stringValue = getStringLiteralValue(value);
+	if (stringValue === 'true') return 'true';
+	if (stringValue === 'false') return 'false';
+
+	return null;
```
AugustinMauroy:
```suggestion
	return getStringLiteralValue(value);
```

AugustinMauroy (changes requested):
there are still comment from my previous review that need be addressed

Herrtian:
Updated the lookup call query to use call_expression/function matching, removed the spread path, applied the shorter guards, and switched the value handling to cached kind/switch checks.

Checked with:
- npm run test --workspace @nodejs/dns-lookup-options-coercion
- npm run type-check
- npm run lint
- git diff --check

recipes/dns-lookup-options-coercion/src/workflow.ts
```diff
@@ -0,0 +1,188 @@
+import { getModuleDependencies } from '@nodejs/codemod-utils/ast-grep/module-dependencies';
+import { resolveBindingPath } from '@nodejs/codemod-utils/ast-grep/resolve-binding-path';
+import type { Edit, SgNode, SgRoot } from '@codemod.com/jssg-types/main';
+import type Js from '@codemod.com/jssg-types/langs/javascript';
+
+const NUMERIC_OPTIONS = new Set(['family', 'hints']);
+const BOOLEAN_OPTIONS = new Set(['all', 'verbatim']);
+const IGNORED_ARGUMENT_KINDS = new Set([',', '(', ')']);
```
AugustinMauroy: I'm pretty sure there are a thing with ast-grep for un-named alias . cc @brunocroh maybe ?

AugustinMauroy (approved):
LGMT!

recipes/dns-lookup-options-coercion/src/workflow.ts
brunocroh: yep you are right, `node.isNamed()` will return false for these already

brunocroh (approved):

avivkeller (approved):

PR #486, opened 2026-06-06
feat(rimraf): add fs.rm migration recipe

### LINKED ISSUE #472: Spec: Rimraf to Native `fs.rm()` Comparison Suite
## Description

We should document and codemod the migration from `rimraf` v3/v4/v5 to `node:fs.rm()` in Node.js v24.

- The migration should be divided by `rimraf` v3, v4, and v5.
- The migration should separate import replacement from runtime behavior differences.
- The migration should call out observed output gaps so version-specific behavior is explicit.
- The migration should keep the deletion wiring and example structure aligned across all baselines.
- The migration should include compensating patterns for code that depends on legacy glob handling or retry behavior.

## Important points

- `rimraf` v3 and v4 expose the older package-style recursive delete APIs, while v5 uses named exports.
- `rimraf` glob patterns do not map directly to `fs.rm()`, but they can be migrated with `fs.globSync()` plus `fs.rm()` or `fs.rmSync()`.
- Most cases are a straight source-import migration, but a few cases need behavior notes because glob handling or missing-path behavior differs at runtime.
- The codemod should prefer built-in `node:fs` imports for Node.js-only code.
- `fs.rmdir({ recursive: true })` is deprecated and should not be used as a replacement.

### rimraf v3 -> node:fs v24

```diff
- import rimraf from "rimraf-v3";
+ import { globSync, rmSync } from "node:fs";
+ import { rm as rmPromise } from "node:fs/promises";
```

Literal recursive delete:

```diff
- rimraf("dist", (error) => {
-   if (error) throw error;
- });
+ await rmPromise("dist", {
+   recursive: true,
+   force: true,
+ });
```

Observed runtime behavior:

```diff
--- rimraf v3
+++ node:fs v24
@@
 (no changes)
```

Example of glob migration:

```diff
- rimraf("dist/**/*.js", (error) => {
-   if (error) throw error;
- });
+ for (const filePath of globSync("dist/**/*.js")) {
+   rmSync(filePath, { recursive: true, force: true });
+ }
```

### rimraf v4 -> node:fs v24

```diff
- import rimraf from "rimraf-v4";
+ import { globSync, rmSync } from "node:fs";
+ import { rm as rmPromise } from "node:fs/promises";
```

Observed runtime behavior:

```diff
--- rimraf v4
+++ node:fs v24
@@
 (no changes)
```

Example of the same migration pattern with no compensation needed:

```diff
- await rimraf("dist", {
-   glob: false,
- });
+ await rmPromise("dist", {
+   recursive: true,
+   force: true,
+ });
```

### rimraf v5 -> node:fs v24

```diff
- import { rimraf, rimrafSync } from "rimraf-v5";
+ import { globSync, rmSync } from "node:fs";
+ import { rm as rmPromise } from "node:fs/promises";
```

Observed runtime behavior:

```diff
--- rimraf v5
+++ node:fs v24
@@
 (no changes)
```

Example of sync migration:

```diff
- rimrafSync("dist");
+ rmSync("dist", {
+   recursive: true,
+   force: true,
+ });
```

## Caveats

- These are migration differences, not stream-style output gaps.
- The version split matters when code depends on glob expansion, missing-path behavior, or Windows retry handling.
- If the code relies on package-specific retry semantics, keep the behavior explicit instead of assuming parity with native deletion.
- If the code only needs literal recursive deletion, prefer the Node core behavior and remove the `rimraf` dependency.
- If the code depends on glob expansion, keep the globbing step separate and use `fs.globSync()` before deletion.

## Refs

- Node.js `fs.rm()` documentation: https://nodejs.org/api/fs.html#fsrmpath-options-callback
- Node.js `fs.globSync()` documentation: https://nodejs.org/api/fs.html#fsglobsyncpattern-options
- Node.js `fs.rmdir()` deprecation notes: https://nodejs.org/api/fs.html#fsrmdirpath-options-callback
- `rimraf` readme: https://github.com/isaacs/rimraf/blob/main/README.md

Herrtian:
Fixes #472.\n\n## Summary\n- add a rimraf-to-fs-rm recipe for literal rimraf deletes\n- handle v3 callback calls, v4/v5 promise calls, v5 sync calls, aliases, and sync glob expansion\n- document manual-review cases for runtime-specific rimraf behavior\n\n## Validation\n- npm.cmd run test --workspace @nodejs/rimraf-to-fs-rm\n- npm.cmd run type-check\n- npm.cmd run lint


AugustinMauroy:
thanks for being volunteering but you have 4 opened PRs we didn't have a lot of maintainer on this initiative so plz don't open any other pr.

AugustinMauroy (changes requested):
* give name to tests cases 
* add remove dep step **but** before calling the function analyse the json ast to know if rimraf isn't used as cli
* add jsdocs to workflow

Herrtian:
Thanks, understood. I won't open more PRs and will focus on the existing ones. I pushed the requested updates in the latest commit.

recipes/rimraf-to-fs-rm/src/remove-dependencies.ts
```diff
@@ -0,0 +1,77 @@
+import type { SgNode, Transform } from '@codemod.com/jssg-types/main';
+import type Json from '@codemod.com/jssg-types/langs/json';
+import removeDependencies from '@nodejs/codemod-utils/remove-dependencies';
+
+const rimrafPackages = [
+	'rimraf',
+	'rimraf-v3',
+	'rimraf-v4',
+	'rimraf-v5',
```
AugustinMauroy: I think nearly nobody use this specifier.

recipes/rimraf-to-fs-rm/src/remove-dependencies.ts
```diff
@@ -0,0 +1,77 @@
...
+	});
+
+	const scriptsObject = scriptsPair?.field('value');
+	if (!scriptsObject) return false;
+
+	const scriptPairs = scriptsObject.findAll({
+		rule: { kind: 'pair' },
+	});
+
+	return scriptPairs.some((pair) => {
```
AugustinMauroy: here I think we can simplify by catching `string_content` kind node inside `string` kind and on `string_content` you can add regex that have rimraf so no need to have js regex ...

recipes/rimraf-to-fs-rm/src/workflow.ts
```diff
@@ -0,0 +1,272 @@
...
+/**
+ * Escapes a binding name so it can be used safely in an ast-grep regex.
+ */
+const escapeRegex = (value: string) =>
+	value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+
+/**
+ * Splits a comma-separated argument list without splitting nested expressions.
+ */
+const splitTopLevelArguments = (source: string): string[] => {
```
AugustinMauroy: @brunocroh I pretty sure that you will have a 3 line function that can simplify that

recipes/rimraf-to-fs-rm/src/workflow.ts
```diff
@@ -0,0 +1,272 @@
...
+			edits.push(call.replace(`rmPromise(${pathArg}, ${buildRmOptions()})`));
+		}
+	}
+
+	if (!edits.length) return null;
+
+	const importReplacement = buildImportReplacement(replacement);
+	rimrafImports.forEach((importNode, index) => {
+		edits.push(importNode.replace(index === 0 ? importReplacement : ''));
+	});
```
AugustinMauroy: use `for(const ... of ...)` with an `if else` statement inside it's will be much more easiest to read

recipes/rimraf-to-fs-rm/src/workflow.ts
```diff
@@ -0,0 +1,272 @@
...
+
```
AugustinMauroy, line 231:
```suggestion

	if (!edits.lenght) return null;
```

for security

recipes/rimraf-to-fs-rm/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Manual review
+
+Some `rimraf` behavior should stay manual because it depends on runtime
+semantics rather than only source syntax:
+
+- custom retry behavior
+- custom glob options
+- async glob deletes with callbacks
+- package-specific fallback behavior on Windows
```
AugustinMauroy:
```suggestion
- package-specific fallback behavior on Windows
- CLI usage isn't touch you may want to use something else
```

recipes/rimraf-to-fs-rm/src/workflow.ts
```diff
@@ -0,0 +1,238 @@
...
+			replacement.usesRmPromise = true;
+			edits.push(call.replace(`rmPromise(${pathArg}, ${buildRmOptions()})`));
+		}
+	}
+
+	const hasCallEdits = edits.length > 0;
+	if (!hasCallEdits) {
+		return null;
+	}
```
AugustinMauroy:
```suggestion
	if (!edits.length ) return null;
```

recipes/rimraf-to-fs-rm/src/workflow.ts
```diff
@@ -0,0 +1,233 @@
...
+};
+
+const RIMRAF_SOURCE_REGEX = '^rimraf(-v[345])?$';
+
+/**
+ * Escapes a binding name so it can be used safely in an ast-grep regex.
+ */
+const escapeRegex = (value: string) =>
+	value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+
```
avivkeller, line 25: `RegExp.escape`?

recipes/rimraf-to-fs-rm/src/workflow.ts
```diff
@@ -0,0 +1,233 @@
...
+const getLineEnding = (source: string) =>
+	source.includes('\r\n') ? '\r\n' : '\n';
```
avivkeller, line 27: IMO just use a `\n`, the users' formatter will handle it from there

recipes/rimraf-to-fs-rm/src/workflow.ts
```diff
@@ -0,0 +1,233 @@
...
+
+	return bindings;
+};
+
+/**
+ * Extracts default and named rimraf bindings from an import statement.
+ */
+const parseImportBindings = (source: string): Binding[] => {
+	const bindings = parseNamedBindings(source);
+	const defaultMatch = source.match(/^import\s+([^,{]+?)(?:\s*,|\s+from\s+)/);
```
avivkeller, line 89: What about dynamic imports?

recipes/rimraf-to-fs-rm/src/workflow.ts
```diff
@@ -0,0 +1,233 @@
...
+		lines.push('import { rm as rmPromise } from "node:fs/promises";');
+	}
+
+	return lines.join(lineEnding);
+};
+
+/**
+ * Builds the recursive rm options that match rimraf's common force behavior.
+ */
+const buildRmOptions = () => '{ recursive: true, force: true }';
```
avivkeller, line 125: THis can be a constant

AugustinMauroy:
@Herrtian any news ?

PR #505, opened 2026-07-13
fix(codemod-utils): handle non-node imports correctly

AugustinMauroy:


utils/src/ast-grep/import-statement.test.ts
```diff
@@ -230,4 +230,42 @@ describe("import-statement", () => {
...
+
+		const ast = astGrep.parse(astGrep.Lang.JavaScript, code);
+
+		const imports = getNodeImportStatements(ast, "test");
+
+		assert.strictEqual(imports.length, 2);
+		assert.deepStrictEqual(
+			imports.map((i) => i.field("source")?.text()),
+			['"node:test"', '"test"'],
+		);
```
JakobJingleheimer: nit: I think these can be combined (where a single deep equal confirms both the contents and the element count).

JakobJingleheimer (commented):
Thanks!

I think we should have some tests with the `node:` prefix and without it.

There are some seemingly unrelated tests added here. Are they actually related? If not, they should go in their own PR.

AugustinMauroy:
> There are some seemingly unrelated tests added here. Are they actually related? If not, they should go in their own PR.

that was to fetch ast and search for type node but it's an unnamed node ...

brunocroh (approved):

alexbit-codemod (approved):

brunocroh (approved):

PR #507, opened 2026-07-13
setup(ci): update + clean

AugustinMauroy:
1. remove version matrix we only need node for npm codemods are run on JSSG
2. un-command `step-security`
3. publish script block list 
4. don't run both push & pull_request


brunocroh (approved):

brunocroh:
The CI is failing with

```md
Post job cleanup.
[harden-runner] post-step
Error: EACCES: permission denied, open '/home/agent/post_event.json'
```

.github/workflows/codemod_publish.yml
```diff
@@ -32,9 +32,16 @@ jobs:
 
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
```
avivkeller, line 35: Why regress?

AugustinMauroy: that git change display 

but previously we had a step that was building an node version matrix. this step had this action but commented dude to issue with `ubuntu-slim`. and when removing it it's do something strange

JakobJingleheimer (approved):

avivkeller (approved):

PR #508, opened 2026-07-14
Fix import path for resolveBindingPath

inamdarsayma-cmyk:


AugustinMauroy:
Wtf is this change it's not a valid js code

JakobJingleheimer:
Reported

PR #512, opened 2026-07-19
feat(v22-to-v24): introduce

AugustinMauroy:
# BOOM !

yeah there are still 3 missing codemod to have fully covered this version but to move forward I pretty sure we need to land that


recipes/v22-to-v24/README.md
```diff
@@ -0,0 +1,12 @@
+# Node.js v22 → v24 Migration
+
+This codemod only wraps all migrations needed to upgrade a Node.js codebase from **v22 to v24**.
+
+> [!WARNING]
+> This codemod is on `v0.0.X` because there are missing parts so be careful before pushing changes to production. If you find any missing migration, please open an issue or a PR.
```
brunocroh, line 6:
```suggestion
> This codemod is on `v0.0.1` because there are missing parts so be careful before pushing changes to production. If you find any missing migration, please open an issue or a PR.
```

brunocroh (approved):

recipes/v22-to-v24/README.md
AugustinMauroy: I didn't do that so when updating the package we don't need to update the readme

avivkeller (approved):

PR #513, opened 2026-07-19
setup(package.json): reduce amout of code to maintain

AugustinMauroy:
## Description

1. reduce amout of thing to maintain on package.json a lot of keys isn't needed since we only publish on codemdo registry 
2. fix some test 
3. fix biome warning


brunocroh (approved):
LGTM

bjohansebas (approved):

